PR #1, opened 2018-08-02
Changed huntergate hash

zsmith3:


zsmith3:
- Added CMakeLists.txt to build adept gmm
- Modified Adept main.cpp to fit new structure (and added some output)
- Changed to x64-debug because Adept/lib/Debug/Adeptd.lib was built in x64

zsmith3:
Manual BA failed, seemingly due to missing Eigen package

PR #2, opened 2018-08-09
Cmake-reorg

zsmith3:
- Ceres and ADOL-C working
- run-all.ps1 working, with new structure
- cmake creates ps1 file with bindir path
- plot_graphs working, with decent 2d graph


ADBench/run-all.ps1
```diff
@@ -1,29 +1,109 @@
-function assert($expr)
-{
-    if (!(& $expr)) {
+# Accept command-line args
+param([int]$global:nruns_f=10, [int]$global:nruns_J=10, [string]$global:tmpdir="")
```
awf: Why do these need to be $global:?

ADBench/run-all.ps1
```diff
@@ -1,29 +1,109 @@
...
         throw "Assertion failed [$expr]"
     }
 }
 
-$dir = split-path ($MyInvocation.MyCommand.Path)
-assert { $dir -match 'ADbench$' }
-$dir = Split-Path $dir
+# Get source dir
+
+$global:dir = split-path ($MyInvocation.MyCommand.Path)
```
awf: Most of these should be not global: -- I can help figure out cases where that breaks something.

ADBench/run-all.ps1
```diff
@@ -1,29 +1,109 @@
...
+	[string]$name
+	[bool]$gmm_both
+	[string]$type
+	[bool]$gmm_use_defs
+
+	# Static constants
 
-Write-Host "$dir"
+	# GMM values - NOTE not all implemented for ceres
+	static [int[]]$gmm_d_vals = @(2, 10) # @(2, 10, 20, 32, 64)
```
awf: May want to store this list somewhere that CMake can read it, then load here.

ADBench/run-all.ps1
```diff
@@ -1,29 +1,109 @@
...
 
-function run($tool, $objective)
-{
-    & "$bindir\tools\$tool\Tools-$tool-$objective.exe" @args
+	# Constructor
+	Tool ([string]$name, [bool]$gmm_both, [string]$type, [bool]$gmm_defs) {
+		$this.name = $name
+		$this.gmm_both = $gmm_both
+		$this.type = $type
+		$this.gmm_use_defs = $gmm_defs
```
awf: Generally keep same name, as you did with the others, i.e. call it gmm_use_defs in the ctor args

ADBench/run-all.ps1
```diff
@@ -1,29 +1,109 @@
...
+# Load bindir variable
+assert "Test-Path" "$global:dir/ADBench/cmake_vars.ps1"
+. $global:dir/ADBench/cmake_vars.ps1
+
+# Set tmpdir default
+if (!$global:tmpdir) { $global:tmpdir = "$dir/tmp" }
+
+
+# Custom Tool class
+Class Tool {
```
awf: Good use of class. 👍

ADBench/run-all.ps1
```diff
@@ -1,29 +1,109 @@
...
+	# Run all tests for this tool
+	[void] runall () {
+		Write-Host $this.name
+		$this.testgmm("")
+	}
+
+	# Run a single test
+	[void] run ([string]$objective, [string]$dir_in, [string]$dir_out, [string]$fn) {
+		$cmd = ""
+		$cmdargs = @($dir_in, $dir_out, $fn, $global:nruns_f, $global:nruns_J)
```
awf: Ah OK, I see why you wanted it global.  You should be able to do $script:nruns_f here.

ADBench/run-all.ps1
zsmith3: It couldn't seem to access them from within the class methods otherwise - said something like 'variable not defined within method'

build adolc.bat
```diff
@@ -0,0 +1,6 @@
+SET currentdir=%cd%
+call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\Tools\VsDevCmd.bat"
+call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\Tools\VsMSBuildCmd.bat"
+cd %currentdir%
+set INCLUDE="%INCLUDE%C:\Users\Zak Smith\Downloads\boost_1_67_0;C:\ZS\autodiff\submodules\adol-c\MSVisualStudio\v14\nosparse;"
```
awf: Can you add a TODO comment wherever there are hard-coded paths.   We can fix them later, but it may be useful to flag them here, as e.g. the Visual Studio ones will be the same on many machines but not all.

cmake/Hunter/config.cmake
```diff
@@ -0,0 +1,4 @@
+hunter_config(glog
```
awf: Comment ceres needs glog to be built with sharedlibs

tools/Ceres/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+# Get Ceres (with GLOG)
+hunter_add_package(ceres-solver)
+find_package(Ceres CONFIG REQUIRED)
+include_directories("${CERES_INCLUDE_DIRS}")
+
+# Create GMM with different #define D and K values
+set(D_VALS 2 10) # NOTE unable to run 20, 32 due to memory error
+set(K_VALS 5 10 25 50) # removed 100, 200 as well
+# NOTE may be able to run more than is done here
+foreach (D_VAL ${D_VALS})
```
awf: Consider emitting these values to a file, to be picked up by run-all.
Also, is the crash at compile time or run time?   If the latter, let it happen -- it means ceres gets a black dot.
If at compile time, we may be able to help it out.

tools/Ceres/main.cpp
```diff
@@ -76,7 +82,7 @@ void compute_gmm_J(int d, int k, int n, double *alphas,
   CostFunction* cost_function =
     new AutoDiffCostFunction<GMMCostFunctor, 1, GMM_K, GMM_D*GMM_K,
     GMM_ICF_DIM*GMM_K>(new GMMCostFunctor(n, x, wishart));
-
+  // NOTE this line (Above)
```
awf: remove comment now?

tools/DiffSharp/build.bat
```diff
@@ -0,0 +1,5 @@
+rem TODO need to call vcvarsall or soemthing
+
+fsc -d:MODE_AD -d:DO_GMM_FULL --lib:"C:\Users\Zak Smith\CMakeBuilds\95e43dd6-1979-0633-8dca-9ab4e04499c8\build\x64-Debug\tools\DiffSharp" --reference:DiffSharp.dll --reference:MathNet.Numerics.dll --reference:MathNet.Numerics.FSharp.dll gmm.fs Program.fs
```
awf: again TODO hard path

ADBench/run-all.ps1
awf: oops, I submitted this comment too early... see fix below.

tools/Ceres/CMakeLists.txt
zsmith3: Crash was compile time

zsmith3: I didn't look into it at all (decided to leave and return to later) so it may be an easy fix - I'll have a look tomorrow.

zsmith3:
adding some unfinished progress in commits, just to back up to GitHub

PR #21, opened 2019-05-10
Toelli/plot

tomjaguarpaw:
Refactorings of `plot_graphs.py`.  Does not change behaviour.


tomjaguarpaw:
If  merging please merge by rebasing!  I've taken great care to split these refactorings into simple commits.

PR #31, opened 2019-05-27
Clean up gitignore

awf:
Now that output files are moved to ``tmp``, can clean up gitignore.


tomjaguarpaw:
Sounds like a good idea.

PR #33, opened 2019-05-27
Clean up the run-all script.

awf:
Comment parameters.  Now ``get-help ADBench/run-all.ps1 -detailed`` gives useful descriptions.
Add "-repeat" option to rerun already-completed tests.
Add "-repeat_failures" to rerun only tests which timed out
Default sampling time to 10 seconds.


ADBench/run-all.ps1
```diff
@@ -181,9 +209,17 @@ Class Tool {
 		elseif ($objective.endswith("SPLIT")) { $out_name = "$($this.name)_split" }
 		else { $out_name = $this.name }
 		$output_file = "${dir_out}${fn}_times_${out_name}.txt"
-		if (!$script:repeat -and (Test-Path $output_file)) {
-			Write-Host "          Skipped test (already completed)"
-			return
+		if (Test-Path $output_file) {
```
tomjaguarpaw: Wouldn't it be clearer to leave `!$script:repeat -and` in the outer `if`?  Nothing is run in the body unless `!$script:repeat`.  The body would become

```powershell
if ($script:repeat_failures) { ... }
else { ... }
```

tomjaguarpaw:
Nice improvements!

ADBench/run-all.ps1
tomjaguarpaw: I can go ahead and do that for you if you like.

awf: Good change, please do!

tomjaguarpaw:
(If you're happy I think we want to squash merge)

PR #37, opened 2019-05-28
Some tweaks for Linux compatibility

tomjaguarpaw:


tomjaguarpaw:
cc @athas

tomjaguarpaw:
(If merging please merge by rebasing)

ADBench/run-all.ps1
```diff
@@ -189,12 +189,7 @@ Class Tool {
 		$cmd = ""
 		$cmdargs = @($dir_in, $dir_out, $fn, $script:nruns_f, $script:nruns_J, $script:time_limit)
 		if ($this.type -eq "bin") {
-			if ($this.name -eq "DiffSharp") {
-				$builddir = ""
-			} else {
-				$builddir = "\$script:buildtype"
-			}
-			$cmd = "$script:bindir\tools\$($this.name)$builddir\Tools-$($this.name)-$objective.exe"
+			$cmd = "$script:bindir\tools\$($this.name)\Tools-$($this.name)-$objective.exe"
```
tomjaguarpaw: @awf: Can you check this still works for you?  On my machine the tools don't get put inside a subdirectory named after their build type.

tomjaguarpaw: Apparently this was addressed by https://github.com/awf/ADBench/pull/38.  It has been closed though I'm not sure why.

ADBench/run-all.ps1
```diff
@@ -237,7 +232,7 @@ Class Tool {
 
 					$dir_in = "$([Tool]::gmm_dir_in)$sz/"
 					$dir_out = "$script:tmpdir/gmm/$sz/$($this.name)/"
-					mkdir -force $dir_out
+					New-Item -Path $dir_out -ItemType Directory -ErrorAction Ignore
```
awf: Is this equivalent to -Force?  I don't think so, as -Force will squeak if it fails.

ADBench/run-all.ps1
```diff
@@ -260,7 +255,7 @@ Class Tool {
 		if ($this.eigen_config[3]) {$objs += @("BA-Eigen") }
 
 		$dir_out = "$script:tmpdir/ba/$($this.name)/"
-		if (!(Test-Path $dir_out)) { mkdir $dir_out }
+		if (!(Test-Path $dir_out)) { New-Item -Path $dir_out -ItemType Directory -ErrorAction Ignore }
```
awf: And here, -force wasn't needed as we already test-pathed it.

ADBench/run-all.ps1
```diff
@@ -189,12 +189,7 @@ Class Tool {
 		$cmd = ""
 		$cmdargs = @($dir_in, $dir_out, $fn, $script:nruns_f, $script:nruns_J, $script:time_limit)
 		if ($this.type -eq "bin") {
-			if ($this.name -eq "DiffSharp") {
-				$builddir = ""
-			} else {
-				$builddir = "\$script:buildtype"
-			}
-			$cmd = "$script:bindir\tools\$($this.name)$builddir\Tools-$($this.name)-$objective.exe"
+			$cmd = "$script:bindir/tools/$($this.name)/Tools-$($this.name)-$objective.exe"
```
awf: Great to get rid of this horror, but would expect to see a change to a fsproj somewhere ensuring the diffsharp exe was put in the right place.

tomjaguarpaw: This change is behaviour-preserving for the diffsharp branch.  It only changes behaviour when `$this.name` is *not* `"DiffSharp"`.

ADBench/run-all.ps1
tomjaguarpaw: I'm not sure what you mean.  I don't think we want to be using `-Force` since the documentation sounds rather scary:

> -Force
>
> Force this cmdlet to create an item that writes over an existing read-only item. 

https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/new-item?view=powershell-6

I think we just want to ignore errors, i.e. when the directory already exists.

ADBench/run-all.ps1
awf: I see, but now it's quite wide-ranging.  The buildtype argument now does very little.  To make a "Debug" graph (not that one really wants one, but it was an option before), one would now need to run cmake in a new build directory, with new default build arguments.

That would really imply that the artifacts should be output in the build directory, not the top-level tmp dir, which itself makes good sense.

ADBench/run-all.ps1
awf: Yes, but not when the directory failed to be made, as we will get error cascades.

I think the only thing to do is the "most correct" thing: make a local
```
function mkdir_p($path) {
   if (test-path -type container $path) 
      return
   else if (test-path $path) 
     error "There's something in the way"
   else 
    new-item -itemtype directory $path
}
```

ADBench/run-all.ps1
tomjaguarpaw: For me the binaries appear in places like `tools/Manual/Tools-Manual-GMM.exe`.  Apparently this is not the case for all versions of VS. @msdmkats has a PR (https://github.com/awf/ADBench/pull/39) to make it the case for all versions.   Would you prefer Release and Debug builds to go in separate subdirectories, i.e.

* `tools/Manual/Release/Tools-Manual-GMM.exe`
* `tools/Manual/Debug/Tools-Manual-GMM.exe`

If so then we will probably need to fix that PR.

msdmkats: It was not a matter of the version of VS, but the cmake's generator used. Some generators, namely Visual Studio/msbuild are multi-configuration generators, i.e. they allow to do builds in different configurations with only one run of cmake, However, most generators (Ninja - the default one, Unix Make) don't have such functionality. And they require separate runs of cmake anyway.

tomjaguarpaw: Thanks for the clarification.  I think it's now up to @awf to decide how to proceed.  If we're trying to support Linux I don't think we'll be able to support this distinction between Debug and Release.

awf: Yes, so I propose to remove "buildtype" from run-all, and make this change.
That will have a further implication that run-all will probably need to be run from inside a cmake build dir, so let's put the "cmake-vars-BUILD.ps1" script into that build dir rather than in the ADbench dir, and then include "./cmake-vars.ps1" in run-all.  If it doesn't exist, then have run-all emit an error saying "This script should be run inside a cmake build directory".

awf: I know it's a maze of twisty passages, but I believe we're removing some of the twists.

ADBench/run-all.ps1
tomjaguarpaw: OK, done

ADBench/run-all.ps1
tomjaguarpaw: OK, I think that can be part of @msdmkats's ongoing work.

tomjaguarpaw: I cancelled the removal of the build type dependent code from this PR.  We can deal with it elsewhere.

athas:
I can confirm that this patch resolves the main problems on Linux (the issues with .exe extensions and Release/Debug directories should be addressed by tweaking cmake, as I understand it).

ADBench/run-all.ps1
tomjaguarpaw: @awf I think we're good to go on this one.  Can you confirm?

awf:
lgtm

PR #40, opened 2019-05-28
Adding Zygote version of BA benchmark

NiklasGustafsson:
This adds a version of the ba_* benchmark using Zygote instead of ForwardDiff.


awf:
Great, thanks!

PR #43, opened 2019-06-01
Draft of Knossos/Futhark-generated GMM.

athas:
This pull request adds a directory `tools/KnossosFuthark` which contains

* `gmm_knossos.fut`: the output of my Knossos-to-Futhark translator applied to the differentiated `gmm.ks`

* `gmm_wrapper.fut`: a small hand-written wrapper that imports `gmm_knossos.fut`, massages input as necessary, and projects out the results that we are interested in.  It defines the entry points `gmm_objective` and `rev_gmm_objective`.

* `gmm_wrapper.c`/`gmm_wrapper.h`: the result of applying the Futhark compiler to `gmm_wrapper.fut`.  This means you don't need to have Futhark installed to use this.

* `gmm.cpp`: a program that reads data and passes it on to the Futhark-generated code.

It's not hooked up to run-all.ps1 yet, nor have I verified whether it actually computes the right Jacobian.  What I'd like is confirmation that I'm moving in the right direction.  (E.g. am I calling the right entry point?  The Knossos wrapper that Tom sent me calls `noderiv_gmm_knossos_gmm_objective_float_reduced2`, which I don't know how relates to the `rev$gmm_knossos_gmm_objective` I'm calling.)


athas:
Turns out it was really easy to do the run-all.ps1 integration.  It's not run by default, however.

Performance is yet unimpressive: n=10k, D=10, K=10, KnossosFuthark runs the objective function in 2.258110e-01s and the derivative in 8.338195e+00s, while Manual does it in 2.996031e-02s and 1.559438e-01s, respectively (more than an order of magnitude faster).

awf:
Thanks for this!  I'll leave @tomjaguarpaw to comment on the entry point, but the Futhark code looks good, as does the run-all.ps1 integration.

tomjaguarpaw:
Yes, you're calling the right entry point.  `noderiv_gmm_knossos_gmm_objective_float_reduced2` is an internal hand-optimized version of `rev$gmm_knossos_gmm_objective` that we use.  In fact I'll provide you with the former when I can so that your benchmark runs faster!

athas:
Some graphs to indicate the current state.  First, performance of KnossosFuthark when run on my Ryzen 1700X CPU.  As I hinted, it's not yet particularly impressive.

And here's a graph when running it on an older K40 GPU (but which has full-speed doubles):

Something is clearly going wrong - I suspect that after a point, memory pressure is making the GPU memory allocator do some trashing.  The Futhark compiler itself is a bit of a memory hog (we're working on it!), and there are also some unfortunate patterns in the Knossos-generated gradient functions.

awf:
Wow, that's the first GPU benchmark I've seen on ADBench.  Kind of amazing that 3 independent variables is worth going to the GPU for; but maybe the 10K datapoints make it worthwhile...  Is the cuda code visible in the PR somewhere?  I couldn't see it in a quick scan of gmm_wrapper.c

athas:
No, I did a local recompile, and it is completely unreadable machine-generated code (and OpenCL at that).  If you're still curious, [this](http://sigkill.dk/junk/gmm_wrapper.c) is the host-level code (which embeds the kernels as a string), and [these](http://sigkill.dk/junk/kernels.cl) are the kernels extracted to a file.  Note that a lot of the code is an artifact of multi-versioning, and may not actually be used for any of the workloads.

athas:
I did some modifications by hand of the Knossos-generated code to remove some of the sparse vectors and matrices.  The intent is to simulate simplifications that I think can feasibly be done by Knossos itself (but which are not realistic for the Futhark compiler).  Then the graph for GPU execution looks like this:

Still not all the way there, but it does start to look promising.

tomjaguarpaw:
Any idea why it stopped at about 2x10^4?  Out of memory, perhaps?

athas:
Yes, it runs out of memory (actually it crashes spectacularly - Futhark gets really flaky near that boundary).

athas:
Here's a graph for n=2.5M, which is enough to saturate the GPU.
GMM (2 5M)  Jacobian  - Release Graph

Still runs out of memory earlier than I'd expect.

awf:
> Yes, it runs out of memory (actually it crashes spectacularly - Futhark gets really flaky near that boundary).
Meaning BSOD or linux equivalent?

athas:
Usually the process segfaults in various ways, but GPU drivers are flaky enough that it can also affect the host machine.  If I've had too many memory errors in the GPU code I run on my home machine, it will freeze if I attempt to suspend it.

athas:
CPU graphs that include Knossos.

1k
10k

awf:
Fantastic, thanks!  @tomjaguarpaw , let's also ensure we have graphs for "non-hand-optimized" knossos, both in its own right, and as an input to Futhark.

tomjaguarpaw:
It's great that Futhark smooths out all those spikes!

PR #44, opened 2019-06-01
Use proper casing when writing timing files.

athas:
Previously, the C++ implementations (at least some of them) would write output files with a different casing than expected by the run-all.ps1 script.  This doesn't matter on case-insensitive file systems, but it does matter on Linux.

I think it's a bit inconsistent that the casing of "manual" depends on whether we are using Eigen or not, but run-all.ps1 specifically lowercases the implementation name in these cases, so I guess it's intentional.

### DIFF
--- a/tools/Manual/main.cpp
+++ b/tools/Manual/main.cpp
@@ -70,13 +70,13 @@ void test_gmm(const string& fn_in, const string& fn_out,
   cout << "err: " << err << endl;
 
 #ifdef DO_CPP
-  string name("manual");
+  string name("Manual");
 #elif defined DO_EIGEN
-  string name("manual_eigen");
+  string name("Manual_Eigen");
 #elif defined DO_EIGEN_VECTOR
-  string name("manual_eigen_vector");
+  string name("Manual_Eigen_Vector");
 #else
-  string name("manual");
+  string name("Manual");
 #endif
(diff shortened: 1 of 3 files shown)


tools/Manual/main.cpp
```diff
@@ -70,13 +70,13 @@ void test_gmm(const string& fn_in, const string& fn_out,
   cout << "err: " << err << endl;
 
 #ifdef DO_CPP
-  string name("manual");
+  string name("Manual");
 #elif defined DO_EIGEN
   string name("manual_eigen");
```
awf: While we're here, let's make these consistent, and remove the special-casing in run-all.

athas: I agree.  I suggest keeping everything capitalised, to match the directory names in `tools`.  I'll update the pull request.

athas: It's a bit unclear what the intended capitalisation for "eigen" and such is.  I'll go with capitalising these as well, as that is what is done by the tool options in run-all.ps1.

awf: Great, thanks.   Feel free to do what feels cleanest.

athas:
I updated the pull request.  Now the naming is consistent (except that "SPLIT" is all-uppercase, but that's related to a different implementation that I cannot test locally anyway).

awf:
Great, thanks.

PR #46, opened 2019-06-03
Make _scandir_rec not choke on directories that contain non-folders.

athas:
Without this change, I get the following error when running `plot_graphs.py`:

```
$ python3 ADBench/plot_graphs.py --save plot
Output directory is: /home/athas/repos/ADBench/tmp/graphs

Traceback (most recent call last):
  File "ADBench/plot_graphs.py", line 46, in <module>
    all_files = [path for path in utils._scandir_rec(in_dir) if "times" in path[-1]]
  File "/home/athas/repos/ADBench/ADBench/utils.py", line 32, in _scandir_rec
    results += [[fn] + file_name for file_name in _scandir_rec(folder + fn, depth + 1)]
  File "/home/athas/repos/ADBench/ADBench/utils.py", line 32, in _scandir_rec
    results += [[fn] + file_name for file_name in _scandir_rec(folder + fn, depth + 1)]
  File "/home/athas/repos/ADBench/ADBench/utils.py", line 29, in _scandir_rec
    if len(os.listdir(folder)) > 0 and os.path.isdir(folder + os.listdir(folder)[0]):
NotADirectoryError: [Errno 20] Not a directory: '/home/athas/repos/ADBench/tmp/graphs/graphs_index.json/'
```

I still don't understand the logic in `_scandir_rec`, which makes decisions based on whether the first element of `os.listdir(folder)` is a directory.  As far as I know, the order of that list is arbitrary.  This patch should at least make it slightly more robust.

### DIFF
--- a/ADBench/utils.py
+++ b/ADBench/utils.py
@@ -23,18 +23,27 @@ def _set_rec(obj, keys, value, append=False):
         return obj
 
 
-# Recursively scan a directory
-def _scandir_rec(folder, depth=0):
+# Recursively scan a directory for files
+#
+# For a directory structure like
+#
+#   a/b1/file1
+#   a/b1/file2
+#   a/b2 (an empty directory)
+#   a/b3/c/file3
+#
+# _scandir_rec("a") yields
+#
+#   ["b1", "file1"]
+#   ["b1", "file2"]
+#   ["b3", "c", "file3"]
+def _scandir_rec(folder):
     folder = folder.rstrip("/") + "/"
-    if len(os.listdir(folder)) > 0 and os.path.isdir(folder + os.listdir(folder)[0]):
-        results = []
-        for fn in os.listdir(folder):
-            results += [[fn] + file_name for file_name in _scandir_rec(folder + fn, depth + 1)]
-    else:
-        results = [[fn] for fn in os.listdir(folder)]
-    # results = [[folder.split("/")[-2]] + fn for fn in results]
-    return results
-
+    for fn in os.listdir(folder):
+        if os.path.isdir(folder + fn):
+            yield from ([fn] + file_name for file_name in _scandir_rec(folder + fn))
+        else:
+            yield [fn]
 
 # Recursively make directories for file/directory
 def _mkdir_if_none(path):


tomjaguarpaw:
I suspect that completely replacing `_scandir_rec` would be better.  Can you give this a try?

```python
def _scandir_rec(folder):
    for dir_, _, files in os.walk(folder):
        for file_ in files:
            yield os.path.join(dir_, file_)
```

Now the `_rec` part of the name doesn't make sense but I think it's better to avoid breaking the API.

awf:
> Now the _rec part of the name doesn't make sense but I think it's better to avoid breaking the API.

But why not make it do the recursive thing?

tomjaguarpaw:
Oh, it does still scan the directory recursively so the name is fine (I guess I was assuming the name related to the fact that it was *implemented* as a recursive function).

athas:
> I suspect that completely replacing `_scandir_rec` would be better. Can you give this a try?
> 
> ```python
> def _scandir_rec(folder):
>     for dir_, _, files in os.walk(folder):
>         for file_ in files:
>             yield os.path.join(dir_, file_)
> ```
> 
> ~Now the `_rec` part of the name doesn't make sense but I think it's better to avoid breaking the API.~

This does not work:

```
$ python3 ADBench/plot_graphs.py --save plot
Output directory is: /home/athas/repos/ADBench/tmp/graphs

Traceback (most recent call last):
  File "ADBench/plot_graphs.py", line 194, in <module>
    print(f"\nPlotted {figure_idx} graphs")
NameError: name 'figure_idx' is not defined
```

This error is a bit misleading, but it occurs because the big loop in `plot_graphs.py` does not run for any iterations, which is ultimately because `_scandir_rec` returns an empty list (at least as far as I can determine).

athas:
Actually, `_scandir_rec` seems to return something sensible, but the conditional in this list comprehension is always False:

```
all_files = [path for path in utils._scandir_rec(in_dir) if "times" in path[-1]]
```

tomjaguarpaw:
Yes, I broke the API.  Let me have another go ...

tomjaguarpaw:
This adheres to the old API.  It's a reasonably sensible function, it works for the current usecases of `_scan_dir` rec and is a strict improvement over the old one in terms of comprehensibility

```python
# Recursively scan a directory for files
#
# For a directory structure like
#
#   a/b1/file1
#   a/b1/file2
#   a/b2 (an empty directory)
#   a/b3/c/file3
#
# _scandir_rec("a") yields
#
#   ["b1", "file1"]
#   ["b1", "file2"]
#   ["b3", "c", "file3"]
def _scandir_rec(folder):
    print("Calling with ", folder)
    folder = folder.rstrip("/") + "/"
    for fn in os.listdir(folder):
        if os.path.isdir(folder + fn):
            yield from ([fn] + file_name for file_name in _scandir_rec2(folder + fn))
        else:
            yield [fn]
```

athas:
That works (after fixing a small typo and removing the debug prints).  I have updated my pull request, but you can also just commit it directly.

tomjaguarpaw:
OK, I submitted it commit it directly.  Thanks for working on this!

PR #48, opened 2019-06-13
Alexn/cppRunner

novikov-alexander:
-Implemented a cpp runner which dynamically loads library containing AD module.
It runs the module according to documentation and writes spent time and results to filesystem.

-Implemented a test which checks whether cpp runner properly loads a module or not.

### DIFF
--- a/src/cpp/runner/main.cpp
+++ b/src/cpp/runner/main.cpp
@@ -1,5 +1,6 @@
 #include <algorithm>
 #include <iostream>
+#include <limits>
 #include <string>
 
 #include "../shared/GMMData.h"
@@ -7,6 +8,7 @@
 #include "../shared/utils.h"
 
 #include "ModuleLoader.h"
+
 
 using namespace std;
 
@@ -29,7 +31,7 @@ inline void call_member_function(unique_ptr<ITest<GMMInput, GMMOutput>>& ptr_to_
 
 double measure_shortest_time(const double minimum_measurable_time, const int nruns, const double time_limit, unique_ptr<ITest<GMMInput, GMMOutput>>& test, const test_member_function func)
 {
-    vector<double> samples;
+    auto min_sample = std::numeric_limits<double>::max();
     double total_time = 0;
     auto repeats = 1;
     while (repeats < (1 << 30))
@@ -40,7 +42,7 @@ double measure_shortest_time(const double minimum_measurable_time, const int nru
         const auto current_run_time = duration_cast<duration<double>>(t2 - t1).count();
         if (current_run_time > minimum_measurable_time)
         {
-            samples.push_back(current_run_time / repeats);
+            min_sample = std::min(min_sample, current_run_time / repeats);
             total_time += current_run_time;
             break;
         }
@@ -53,11 +55,9 @@ double measure_shortest_time(const double minimum_measurable_time, const int nru
         call_member_function(test, func, repeats);
         auto t2 = high_resolution_clock::now();
         const auto current_run_time = duration_cast<duration<double>>(t2 - t1).count();
-        samples.push_back(current_run_time / repeats);
+        min_sample = std::min(min_sample, current_run_time / repeats);
         total_time += current_run_time;
     }
-
-    const auto min_sample = *std::min_element(std::begin(samples), std::end(samples));
 
     return min_sample;
 }
--- a/src/cpp/runner/main.cpp
+++ b/src/cpp/runner/main.cpp
@@ -9,9 +9,6 @@
 
 #include "ModuleLoader.h"
 
-
-using namespace std;
-
 GMMInput read_gmm_data(const string& input_file, const bool replicate_point)
 {
     GMMInput input;
--- a/src/cpp/runner/main.cpp
+++ b/src/cpp/runner/main.cpp
@@ -1,3 +1,4 @@
+#include <algorithm>
 #include <iostream>
 #include <string>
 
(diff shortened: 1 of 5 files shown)


src/cpp/runner/main.cpp
```diff
@@ -9,26 +10,144 @@
...
+    {
+        auto t1 = high_resolution_clock::now();
+        call_member_function(test, func, repeats);
+        auto t2 = high_resolution_clock::now();
+        const auto current_run_time = duration_cast<duration<double>>(t2 - t1).count();
+        samples.push_back(current_run_time / repeats);
+        total_time += current_run_time;
+    }
+
+    const auto min_sample = *std::min_element(std::begin(samples), std::end(samples));
```
msdmkats: We are computing the minimum sample and and the total time. Both of these can be computed by folding, so we don't really need a vector here.

src/cpp/runner/main.cpp
```diff
@@ -9,26 +10,144 @@
...
+
+    return min_sample;
+}
+
+std::string filepath_to_basename(const std::string& filepath)
+{
+    const auto filename = filepath.substr(filepath.find_last_of("/\\") + 1);
+    const auto dot = filename.find_last_of('.');
+    auto basename = filename.substr(0, dot);
```
msdmkats: What happens if there's no dot in the file name?

src/cpp/runner/main.cpp
```diff
@@ -9,26 +10,144 @@
...
-int main(int argc, const char* argv[])
+int main(const int argc, const char* argv[])
 {
-	if (argc < 7) {
-		std::cerr << "usage: CPPRunner modulePath dir_in dir_out file_basename nruns_F nruns_J [-rep]\n";
-		return 1;
-	}
+    try {
+        if (argc < 8) {
+            std::cerr << "usage: CPPRunner module_path input_filepath output_dir minimum_measurable_time nruns_F nruns_J time_limit [-rep]\n";
```
msdmkats: We also need to differentiate between benchmarks (GMM, BA, HAND, HAND_COMPLICATED, LSTM)

src/cpp/runner/main.cpp
```diff
@@ -9,26 +10,144 @@
...
+        const auto time_limit = std::stod(argv[7]);
+
+        // read only 1 point and replicate it?
+        const auto replicate_point = (argc > 8 && string(argv[8]) == "-rep");
+
+        ModuleLoader module_loader(module_path);
+        auto test = module_loader.GetTest();
+
+        auto inputs = read_input_data(input_filepath, replicate_point);
+        auto gradient_size = (inputs.k * (inputs.d + 1) * (inputs.d + 2)) / 2;
```
msdmkats: This size is a part of the output vector. No need to compute it separately here.

src/cpp/runner/main.cpp
```diff
@@ -9,26 +10,144 @@
...
-	auto inputs = readInputData();*/
+        auto input_basename = filepath_to_basename(input_filepath);
+        auto module_basename = filepath_to_basename(module_path);
 
-	test->output();
-}
\ No newline at end of file
+        save_time_to_file(output_dir + input_basename + "_times_" + module_basename + ".txt", objective_time, derivative_time);
+        save_objective_to_file(output_dir + input_basename + "_F_" + module_basename + ".txt", objective_time);
+        save_gradient_to_file(output_dir + input_basename + "_J_" + module_basename + ".txt", gradient_size, output.gradient.data());
```
msdmkats: This code expects `output_dir` to have a trailing slash. We need to ensure, it's there.

src/cpp/runner/main.cpp
```diff
@@ -9,26 +10,144 @@
...
+    // Read instance
+    read_gmm_instance(input_file, &input.d, &input.k, &input.n,
+        input.alphas, input.means, input.icf, input.x, input.wishart, replicate_point);
+
+    return input;
+}
+
+typedef  void (ITest<GMMInput, GMMOutput>::* test_member_function) (int);
+
+inline void call_member_function(unique_ptr<ITest<GMMInput, GMMOutput>>& ptr_to_object, const test_member_function ptr_to_member, const int arg1) {
```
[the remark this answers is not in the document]
novikov-alexander: There is said [here ](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es31-dont-use-macros-for-constants-or-functions )that we should avoid macros as much as possible. 
So, I propose a template.

src/cpp/runner/main.cpp
novikov-alexander: It will be on a next step of development :-)

src/cpp/runner/main.cpp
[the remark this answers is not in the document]
msdmkats: Can a template help with enforcing inlining? I agree, that we should avoid macros as much as possible. I argue, that here is the place where we actually need one. Unless, we are willing to go with the direct C++ syntax.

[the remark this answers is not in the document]
awf: Given how "repeats" works, any overhead for calling this will not affect timing, so it's fine to simply remove the "inline".   (It will with very high probability be inlined anyway).

src/cpp/runner/main.cpp
```diff
@@ -9,26 +10,144 @@
...
+
+    return basename;
+}
 
-}*/
+void save_time_to_file(const string& filepath, const double objective_time, const double derivative_time)
+{
+    std::ofstream out(filepath);
+    out << std::scientific << objective_time << "\t" << derivative_time;
+    out.close();
```
awf: Probably worth adding a newline, as it is human readable anyway, so it's reasonable to make it human-pleasant.  E.g. I sometimes ``cat gmm_times.txt`` on the commandline.

src/cpp/runner/main.cpp
awf: Or just call the variable ``output_prefix``, then if it ends in a slash, it will name a directory, otherwise just a filename prefix.

src/cpp/runner/main.cpp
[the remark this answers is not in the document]
msdmkats: Yes, you're right. Since this will be called only once per time measurement (and not `repeats` times) it won't produce any significant effect on the benchmark result. Probably, we could even have gotten away with std::function/bind. No reason to do it now, that we've got pointers working, though.

PR #50, opened 2019-06-21
CPPRunner which supports both GMM and BA problems and allows to add new problems easily.

novikov-alexander:
- C++ runner checks parameter "test_type" to understand which test it should to run
- C++ is able to run modules with BA tests
- all benchmark functionality is templated and moved to "Benchmark.h" and besides all specialized implementation of benchmark are moved to corresponding files ("GMMBenchmark.h", "GMMBenchmark.cpp" etc.)


src/cpp/runner/main.cpp
```diff
@@ -1,151 +1,44 @@
...
-
-    for (const auto& i : gradient)
-    {
-        out << std::scientific << i << std::endl;
-    }
-
-    out.close();
+std::string str_toupper(std::string s) {
+    std::transform(s.begin(), s.end(), s.begin(),
+        [](const unsigned char c) { return std::toupper(c); }
```
awf: Is it the case that std::toupper is still a macro?  Or wrong signature?

test/cpp/modules/manual/BaTests.cpp
```diff
@@ -8,7 +8,7 @@ TEST(BaTests, Load) {
 #else
     ModuleLoader moduleLoader("../../../../src/cpp/modules/manual/Manual.dll");
 #endif
-    auto test = moduleLoader.GetBaTest();
+    auto test = moduleLoader.get_ba_test();
```
awf: Yes, I prefer this.

test/cpp/modules/manual/BaTests.cpp
```diff
@@ -19,7 +19,7 @@ TEST(BaTests, TestProcess)
 #else
     ModuleLoader moduleLoader("../../../../src/cpp/modules/manual/Manual.dll");
```
awf: Ultimately this to not be hard coded I assume?  But another PR?

msdmkats: The plan is to have (in near future) a parametrized test suite with paths to all dlls as a set of parameters. So that all the ITest<I, O> implementations could be tested simultaneously. But yes, in another PR.

src/cpp/runner/main.cpp
novikov-alexander: Sorry I don't understand what do you mean Andrew. Could you explain more?

PR #52, opened 2019-06-26
Tests of corresponding methodology and c++ runner

novikov-alexander:
-Time always passes as duration<double> instead double. It helps to understand that time measures in seconds.
-Implemented a test which checks benchmark interrupt after exceeding the timelimit.
-Implemented a test which checks benchmark interrupt after exceeding the run limit. Moreover it checks that benchmarking loop is not optimized.
-Implemented a test which checks the function calculating repeats necessary for minimum measurement time archivement.
-Implemented a test which checks that time measured correctly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,9 +64,10 @@ add_subdirectory ("tools")
 option(ENABLE_TESTS "Enable tests" ON)
 
 if (${ENABLE_TESTS})
-    # Get gtest package
+    # Get gtest and gmock packages
     hunter_add_package(GTest)
-    find_package(GTest CONFIG REQUIRED)
+	
+	find_package (GTest CONFIG REQUIRED)
     find_package(GMock CONFIG REQUIRED)
 
     if(NOT EXISTS "${GTest_LICENSES}")
--- a/src/cpp/runner/Benchmark.h
+++ b/src/cpp/runner/Benchmark.h
@@ -85,7 +85,7 @@ duration<double> measure_shortest_time(const duration<double> minimum_measurable
 {
     auto find_repeats_result = find_repeats_for_minimum_measurable_time(minimum_measurable_time, test, func);
 
-    if (find_repeats_result == measurable_time_not_achieved)
+    if (find_repeats_result.repeats == measurable_time_not_achieved)
     {
         throw exception("It was not possible to reach the number of repeats sufficient to achieve the minimum measurable time.");
     }
--- a/src/cpp/runner/Benchmark.h
+++ b/src/cpp/runner/Benchmark.h
@@ -9,6 +9,7 @@
 using std::chrono::duration;
 using std::chrono::duration_cast;
 using std::chrono::high_resolution_clock;
+using std::chrono::seconds;
 
 /*
  * General logic of the benchmark described in this file.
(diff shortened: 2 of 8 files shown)


test/cpp/runner/CppRunnerTests.cpp
```diff
@@ -1,12 +1,84 @@
...
+}
+
+TEST(CppRunnerTests, TimeLimit) {
+
+    auto gmm_test = module_loader.get_gmm_test();
+    const auto minimum_measurable_time = 0s;
+    const auto run_count = 100; //Run count guarantees total time greater than the time_limit
+    const auto time_limit = 0.1s;
+    const auto execution_time = 0.01s;
+
```
msdmkats: Please, add a comment about `execution_time` having to be greater than `minimum_measurable_time` for test assertions to be correct. (we need this so that `find_repeats...` won't do extra calls)

test/cpp/runner/CppRunnerTests.cpp
```diff
@@ -1,12 +1,84 @@
...
+    auto gmm_test = module_loader.get_gmm_test();
+    const auto assumed_repeats = 16;
+    const auto minimum_measurable_time = 0.01s;
+    
+    auto min_sample = duration<double>(seconds(std::numeric_limits<seconds::rep>::max()));
+    auto total_time = duration<double>(0s);
+    auto repeats = find_repeats_for_minimum_measurable_time(minimum_measurable_time, gmm_test,
+                                                            &ITest<GMMInput, GMMOutput>::calculateObjective, min_sample,
+                                                            total_time);
```
msdmkats: Am I missing something, or `repeats` will evaluate to `1 << 30`, because calculateObjective does nothing?

CMakeLists.txt
```diff
@@ -64,9 +64,10 @@ add_subdirectory ("tools")
 option(ENABLE_TESTS "Enable tests" ON)
 
 if (${ENABLE_TESTS})
-    # Get gtest package
+    # Get gtest and gmock packages
```
awf: Great, comment fixes always welcome!

src/cpp/runner/Benchmark.h
```diff
@@ -37,35 +37,48 @@ void call_member_function(std::unique_ptr<ITest<Input, Output>>& ptr_to_object,
 template<class Input, class Output>
 unique_ptr<ITest<Input, Output>> get_test(const ModuleLoader& module_loader) = delete;
 
-//Measures time according to the documentation.
 template<class Input, class Output>
-double measure_shortest_time(const double minimum_measurable_time, const int nruns, const double time_limit, std::unique_ptr<ITest<Input, Output>>& test, const test_member_function<Input, Output> func)
+int find_repeats_for_minimum_measurable_time(const duration<double> minimum_measurable_time,
+    std::unique_ptr<ITest<Input, Output>>& test,
+    const test_member_function<Input, Output> func, duration<double>& min_sample,
```
awf: Are these reference args output args?  If so, please pass by pointer, else by value or const ref.

test/cpp/runner/CppRunnerTests.cpp
novikov-alexander: Thanks! It is a great idea for another one test.

src/cpp/runner/Benchmark.h
novikov-alexander: I've replaced it by returning a struct.

PR #54, opened 2019-07-01
Alexn/cpp runner hand

novikov-alexander:


msdmkats:
Am I correct that you rebased this on top of dmkats/manulaHand?

novikov-alexander:
> Am I correct that you rebased this on top of dmkats/manulaHand?
Yes, you are right

PR #59, opened 2019-07-10
CppRunner Linux support

novikov-alexander:
-CppRunner compiles under Linux
-RepeatsNotFound test fixed
-TimeMeasurement test fixed
-references to unique_ptr's replaced by usual references
-identical extensions for Linux and Windows binaries

### DIFF
--- a/src/cpp/shared/ITest.h
+++ b/src/cpp/shared/ITest.h
@@ -1,10 +1,19 @@
 #pragma once 
 
-#if __GNUC__ >= 4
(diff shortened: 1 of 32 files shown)


src/cpp/shared/ITest.h
```diff
@@ -1,6 +1,11 @@
 #pragma once 
 
-#include "defs.h"
+#if __GNUC__ >= 4
```
msdmkats: This will work incorrectly when compiler is neither cl nor gcc v4+, as it will resolve to `__declspec` which is cl-specific.

PR #65, opened 2019-08-06
Finite differences c++ module

msdmkats:
- Implemented finite differences as a module for CppRunner
- Tested it
- Corrected jacobian computation for hand tracking (used to compute only one pair of derivatives by us)
- Optimized some memory allocations

Some of the resulting plots:
FiniteOld here stands for the old (removed in this PR) implementation.

GMM (1k)  Jacobian  - Release Graph
Here new implementation works longer than the old one, because the old one didn't compute all the derivatives by us.
BA  Jacobian  - Release Graph
LSTM  Jacobian  - Release Graph


tomjaguarpaw:
I'm not an expert on this, but I can't see anything wrong with it!

src/cpp/modules/finite/finite.h
```diff
@@ -0,0 +1,113 @@
...
+    ///		Will be stored as a vector(input_size * output_size).
+    ///		Will store in format foreach (input) { foreach (output) {} }</param>
+    /// <param name="delta">The difference to use in finite differentiation</param>
+    void finite_differences_continue(std::function<void(T*, T*)> func,
+        T* input, int input_size, const T* output, int output_size,
+        T* result, T delta = FINITE_DIFFERENCES_DEFAULT_DELTA)
+    {
+        for (int i = 0; i < input_size; i++)
+        {
+            input[i] += delta;
```
awf: Suggest this in case (a + b) - b can not equal a
```
auto tmp = input[i];
input[i]+=delta;
…
input[i] = tmp;
```

msdmkats: Yes, I did only performance tweaks for the original finite differences code without actually checking the correctness.
Please, see https://github.com/awf/ADBench/pull/68 that fixes this as well as a couple of other issues.

PR #67, opened 2019-08-08
Modules API Pascal and camel to snake case + Linux build fix

novikov-alexander:
- names of external dll functions are converted into snake_case
- names of c++ modules API functions are converted into snake_case

### DIFF
--- a/src/cpp/shared/HandEigenData.h
+++ b/src/cpp/shared/HandEigenData.h
@@ -1,6 +1,6 @@
 #pragma once 
 
-#include "../../shared/hand_eigen_model.h"
+#include "./hand_eigen_model.h"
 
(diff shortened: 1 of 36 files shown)


src/cpp/shared/HandEigenData.h
```diff
@@ -1,6 +1,6 @@
 #pragma once 
 
-#include "../../shared/hand_eigen_model.h"
+#include "./hand_eigen_model.h"
```
awf: This is separate to the big rename?  Add to PR title?

novikov-alexander: I've just found out that it began to break Linux build at some moment. That's why I fixed it.

PR #68, opened 2019-08-09
Improved accuracy for finite differences.

msdmkats:
- Made sure that the step in finite differences is comparable to the input variable
- Made sure that finite differences won't modify its input

### DIFF
--- a/src/cpp/modules/finite/finite.h
+++ b/src/cpp/modules/finite/finite.h
@@ -119,7 +119,7 @@ public:
             tmp_f = tmp_b + dx;
             // adjusting dx so that (tmp_b + dx) - tmp_b == delta
             dx = tmp_f - tmp_b;
-            if (dx < delta * 0.02)
+            if (dx < delta * 0.5)
                 std::cerr << "WARNING: Finite difference step " << delta << " seems incompatible with the argument " << originalInput << std::endl;
 
             input[i] = tmp_f;
--- a/src/cpp/modules/finite/finite.h
+++ b/src/cpp/modules/finite/finite.h
@@ -3,13 +3,35 @@
 #include <vector>
 #include <functional>
 #include <algorithm>
+#include <iostream>
 
 const double FINITE_DIFFERENCES_DEFAULT_EPSILON = std::cbrt(std::numeric_limits<double>::epsilon());
 const double FINITE_DIFFERENCES_DEFAULT_ZERO_NEIGHBORHOOD_RADIUS = 1;
 const double FINITE_DIFFERENCES_DEFAULT_ZERO_NEIGHBORHOOD_CHARACTERISTIC_SCALE = 1;
 
+/// <summary>Generates a function that returns a value proportional to some characteristic scale
+///     when the argument lies within a neighborhood of zero or a value proportional to the
+///     argument itself otherwise.</summary>
+/// <param name="epsilon">Coefficient by which the value of the input variable is multiplied
+///     to determine the difference to use in finite differentiation when the absolute value
+///     of the said variable is greater or equal to zero_neighborhood_radius</param>
+/// <param name="zero_neighborhood_radius">Radius of the neighborhood of zero where the difference
+///     used in finite differentiation should not be proportional to the input</param>
+/// <param name="zero_neighborhood_characteristic_scale">Value, which when multiplied by epsilon
+///     gives the difference to use in finite differentiation when the absolute value
+///     of the said variable is lesser than zero_neighborhood_radius</param>
+template<typename T>
+std::function<T(T)> finite_differences_default_step_function(T epsilon = FINITE_DIFFERENCES_DEFAULT_EPSILON, T zero_neighborhood_radius = FINITE_DIFFERENCES_DEFAULT_ZERO_NEIGHBORHOOD_RADIUS, T zero_neighborhood_characteristic_scale = FINITE_DIFFERENCES_DEFAULT_ZERO_NEIGHBORHOOD_CHARACTERISTIC_SCALE) {
+    return [epsilon, zero_neighborhood_radius, zero_neighborhood_characteristic_scale](T x) {
+        T absx = std::abs(x);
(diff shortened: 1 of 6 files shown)


msdmkats:
Ok, I created a separate PR for input preservation here: https://github.com/awf/ADBench/pull/71

As for this change - I stand by it. when we do forward difference, using relative step (outside of a neighborhood of zero, where it can lead to underflow), makes sense - we have a finite significand - and is a standard practice. E.g. Ceres, you mentioned, does that - http://www.ceres-solver.org/numerical_derivatives.html#implementation-details

The reason, we do forward difference and not Riddler's (or at least central difference) is the time to compute - seems to be long as is. If you don't mind multiplying it by a factor we can consider a more accurate scheme.

msdmkats:
After some experiments, I'd like to propose switching to central differences, because forward difference is too inaccurate in our cases. E.g. LSTM loss function's gradient computed with forward difference when the input sequence is 4096 symbols long has almost nothing to do with the actual one computed manually. When I used central differences almost all discrepancies between it and manual went away. There're still some problems with GMM, when there're 2.5M data points, so at the moment I'm trying to compute this gradient using some other tool to check which module is wrong, but otherwise central differences seem good enough for our tasks.
I'm still reluctant to use a more complex difference scheme because, as I said above, runtimes a high already, but central differences seem a good compromise to me.
@awf what do you think?

msdmkats:
Checked gmm/2.5M/d2k10 with Julia forward diff - manual is correct.

awf:
I agree that central is best for comparing manual to.  I am still on the fence about relative steps -- my feeling is that if one's function is not linear in the 1e-6 region, a bunch of other stuff will go wrong anyway.  I know others do it (e.g. Ceres), but...  let me give it another day's thought.  Perhaps @timhutton has an opinion?

timhutton:
@awf Yes, I've seen no benefit of more complicated finite difference schemes when I've tried it. Unless there are discontinuities.

msdmkats:
A couple of comments from me:
1. In the vicinity of zero relative step is not used and the code explicitly falls back to an absolute one - too small step leads to even bigger errors, because of the catastrophic cancellation in the finite difference formula. Relative step is needed for large arguments.
1. I skimmed through our data, and it appears, that number bigger than 2 are quite rare, and there were no (at least I didn't find any) numbers larger than 1000. In that light, not having a relative step would not be _that_ problematic, provided, of course, we'll never introduce bigger values.

msdmkats:
So, @awf as we discussed in a call, `FiniteDifferencesEngine` now accepts a callback to compute the step, with the default one being what I implemented initially.

I also investigated, how zero neighborhood radius/characteristic scale affects accuracy. I picked a few problem of every kind and ran for them manual, finite with constant step of 1e-6, and finite with `epsilon = 1e-6` and  zero neighborhood radius and characteristic scale simultaneously equal to 1e-6, 1e-4, 1e-2, and 1 (also tried bigger ones, but they simply lead to sub-optimally big step in zero neighborhood). Then, for all finite runs I computed maximum and average error as defined in #76 

Here are the results.

Raw:
[avgdlogclean.txt](https://github.com/awf/ADBench/files/3616699/avgdlogclean.txt)
[maxdlogclean.txt](https://github.com/awf/ADBench/files/3616700/maxdlogclean.txt)

Plots (FiniteX - finite with radius/scale of 1eX, FiniteC - fixed step 1e-6):

Max error:
MaxErrorAll
MaxErrorWOTop
MaxErrorWOLeft

Average error:
AvgErrorAll
AvgErrorWOTop
AvgErrorWOLeft

So, zero neighborhood radius and characteristic scale of 1.0 look like a good choice to me.

src/cpp/modules/finite/finite.h
```diff
@@ -63,51 +93,41 @@ class FiniteDifferencesEngine
...
-            func(input, tmp_output.data());
-            div_vec(sub_vec(tmp_output.data(), output, output_size), output_size, delta);
-            vec_ins(&result[output_size * i], tmp_output.data(), output_size);
+            T delta = step(originalInput);
+            tmp_b = originalInput - delta;
+            T dx = delta * 2;
+            tmp_f = tmp_b + dx;
+            // adjusting dx so that (tmp_b + dx) - tmp_b == delta
+            dx = tmp_f - tmp_b;
+            if (dx < delta * 0.02)
```
awf: Did you consider 0.5?  Does it generate more errors?
I'm kind of keen on powers of two for this constant, even though I realize it's somewhat arbitrary.

src/cpp/modules/finite/finite.h
```diff
@@ -2,16 +2,42 @@
...
 #include <functional>
+#include <algorithm>
+#include <iostream>
 
-const double FINITE_DIFFERENCES_DEFAULT_DELTA = 1e-7;
+const double FINITE_DIFFERENCES_DEFAULT_EPSILON = std::cbrt(std::numeric_limits<double>::epsilon());
+const double FINITE_DIFFERENCES_DEFAULT_ZERO_NEIGHBORHOOD_RADIUS = 1;
+const double FINITE_DIFFERENCES_DEFAULT_ZERO_NEIGHBORHOOD_CHARACTERISTIC_SCALE = 1;
+
+/// <summary>Generates a function that returns a value proportional to some characteristic scale
```
awf: Love it, thanks.

awf:
Great investigation, thanks.  If it's relatively easy, you would make your point even more convincing by using log y scale on the error graphs, but this is great.

msdmkats:
Graphs with log scale for y (removed plots with zeroes):

MaxErrorLog
AvgErrorLog

src/cpp/modules/finite/finite.h
msdmkats: No, not initially.

No it doesn't - neither of the two generated any errors so far.

To think of it, for `dx` to be less that `0.5 * delta`, while without round-off errors it would be `2 * delta`, it, probably, must be zero, because that means that the round-off error of the last addition was greater than `0.5 * dx`, so not a single bit of `dx`'s significand was added correctly.

So, yeah, let's make it 0.5.

PR #69, opened 2019-08-09
Need to include <cctype> for std::toupper and std::isalnum

tomjaguarpaw:
I had to include `<cctype>` to get the C++ files to build.  I would be interested to learn how everyone else managed to build them without this include.  Is it perhaps a question of Visual C++ version?

### DIFF
--- a/src/cpp/runner/main.cpp
+++ b/src/cpp/runner/main.cpp
@@ -1,5 +1,6 @@
 ﻿#include <iostream>
 #include <string>
+#include <cctype>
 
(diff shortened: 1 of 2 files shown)


tomjaguarpaw:
I am using

> Microsoft (R) C/C++ Optimizing Compiler Version 19.16.27031.1 for x64

I'm not sure what that corresponds to, but `19.xxx` sounds like VS2019.  It's strange.  Anyway, I shall merge it.

PR #72, opened 2019-08-19
Manual Eigen lstm module

iliaeg:
- Implemented lstm eigen module


iliaeg:
LSTM  Jacobian  - Release Graph
LSTM  Jacobian ÷ objective  - Release Graph
LSTM  Objective  - Release Graph

iliaeg:
- [ ] Tests on linux passed

src/cpp/modules/manualEigen/lstm_eigen_d_structures.h
```diff
@@ -0,0 +1,274 @@
...
+
+#include <vector>
+
+#include "../../shared/lstm_eigen_types.h"
+
+template<typename T>
+struct StateElementGradientModel
+{
+    // Matrix(hsize, 10) derivatives from all layers
+    MapX10<T> d_rawX10;
```
msdmkats: Here and below `MapXN` varaiables need explanations in comments: what exactly are they (which derivatives they incorporate) and why do we need them.

src/cpp/modules/manualEigen/lstm_eigen_d.cpp
```diff
@@ -0,0 +1,366 @@
...
+    state.cell = orig_cell * forget + ingate * change;
+    // wrt weight
+    jacobian.cell.d_weight_forget = orig_cell * forget_dw;
+    jacobian.cell.d_weight_ingate = change * ingate_dw;
+    jacobian.cell.d_weight_outgate = 0.;
+    jacobian.cell.d_weight_change = ingate * change_dw;
+    // wrt bias
+    jacobian.cell.d_bias_forget = orig_cell * forget_db;
+    jacobian.cell.d_bias_ingate = change * ingate_db;
+    jacobian.cell.d_bias_outgate.setZero();
```
msdmkats: Why do we use use `= 0.` for `d_weight_outgate` and `.setZero()` for `d_bias_outgate`? Is there any difference?

iliaeg: No, its not. I've just overlooked that.
Fixed now.

src/cpp/modules/manualEigen/lstm_eigen_d_structures.h
iliaeg: Comments added.

PR #76, opened 2019-08-21
Jacobian correctness checking

msdmkats:
__NB__ This branch if based on dmkats/finitefix branch, so the corresponding PR #68 should be resolved and merged before this. Corollary: discussion of the changes to `src/cpp/modules/**/` should happen there as well.

This change proposes and implements a method to check the accuracy of the Jacobians produced by different AD tools.

`/docs/JacobianCheck.md` describes the methodology.
`/src/cpp/utils/finitePartialGmm` - cpp console app to compute a part of GMM gradient using finite differences
`/src/dotnet/utils/JacobianComparisonLib` - a .NET Standard library (so callable from Powershell[Core]) for comparing Jacobians.
`/ADBench/compare-manvfinite.ps1` - script to compare Jacobians produced by `Manual` and `Finite` to ensure the correctness of `Manual` (I'm running that for days now).
`/ADBench/run-all.ps1` - now uses JacobianComparisonLib to compare results produced by tools other than `Manual` to those produced by `Manual` and outputs the results in a JSON file near the file with timings


docs/JacobianCheck.md
```diff
@@ -0,0 +1,58 @@
+# Jacobian Correctness Verification
+
+ADBench needs not olny to measure the time different tools take to compute the Jacobians of objective functions, but also to check, whether produced Jacobians are accurate. This document describes, how the accuracy of the Jacobians is verified.
+
+## Source of the Ground Truth
+
+Some of our tests (most notably, those related to bundle adjustments) result in very large Jacobians. In fact, a complete set of correct Jacobians for all our tests in tsv format consumes more than 20 GB of storage. As such, it is not feasible to check such a set into the repository. So instead, we designate one of the _testing modules_ to be _golden module_ - the source of ground truth, against output of which Jacobians produced by all the other _testing modules_ will be checked.
```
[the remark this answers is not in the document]
msdmkats: If the Jacobians consumed less memory, we could check them into the repository. That way we wouldn't need a golden module.

ADBench/run-all.ps1
```diff
@@ -283,6 +328,14 @@ Class Tool {
         if (!(test-path $output_file)) {
             throw "Command ran, but did not produce output file [$output_file]"
         }
+
+        if ($this.check_results -and (![string]::IsNullOrEmpty([Tool]::golden_tool_name))) {
```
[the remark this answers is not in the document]
msdmkats: Yes. However, that won't provide much speedup. Comparison is pretty efficient now.

msdmkats:
I finally completed comparing `Manual` and `Finite`.  TSV log is attached here:

[manvfinitelog.txt](https://github.com/awf/ADBench/files/3566409/manvfinitelog.txt)

Seems fine to me.

tomjaguarpaw:
The Jacobian correctness checking doc seems excellent.  I opened this for some small tweaks: https://github.com/awf/ADBench/pull/90

docs/Architecture.md
```diff
@@ -76,7 +76,7 @@ _Benchmark runners_ output 3 files into the folder specified by the _global runn
 
 ### Global Runner
 
-_Global runner_ is a script that consequitively runs all benchmarks using corresponding _benchmark runners_ while enforcing specified hard time limits.
+_Global runner_ is a script that consecutively runs all benchmarks using corresponding _benchmark runners_ while enforcing specified hard time limits.
```
awf: One day, this might run benchmarks in parallel, so suggest simply to delete "consecutively" -- it is not necessary to understand this methodology?

docs/JacobianCheck.md
```diff
@@ -0,0 +1,58 @@
...
+
+The accuracy of the results produced by the _golden module_ is verified separately.
+
+We designate the module `Manual`, which computes the Jacobians using the manually derived formulae, as the _golden module_. We do so, because
+- It is fast enough not to timeout, so when some other module produces a Jacobian, we would have a Jacobian, we consider accurate, to compare it against
+- We have the complete control over this module, so if we ever encounter a bug in it, it will be within our power to fix it
```
awf: Good points!

docs/JacobianCheck.md
```diff
@@ -0,0 +1,58 @@
...
+        </code>
+    </td>
+    <td>(1)</td>
+  </tr>
+</table>
+
+is smaller than the given _tolerance_. Note, that the formula (1) produces the absolute difference between `x` and `y`, when both lie in the vicinity of zero, and the relative one otherwise, which is a suitable way of comparing floating-point numbers.
+
+We generally use the tolerance of `1e-6` to allow for rounding errors during both the computation and the output. This number may be adjusted for specific tests, e.g. it can be increased when the _testing module_ uses single-precision arithmetic.
+
```
awf: 1e-6 is quite large for double -- have you tried 1e-8?

docs/JacobianCheck.md
```diff
@@ -0,0 +1,58 @@
...
+- Maximum encountered difference (1) between the corresponding elements of the Jacobians
+- Average difference (1) between the corresponding elements of the Jacobians
+- Number of encountered differences (1) that exceed the tolerance (`1e-6`)
+- Total number of the compared pairs of elements
+
+Note, that we should expect that some differences between the corresponding elements of the Jacobian will be larger than the tolerance, because of the central difference formula's truncation error.
+This happens especially often for the 2.5M points GMM problem.
+
+So, if the `Manual` module is implemented incorrectly, we would immediately see that reflected in the log in the form large average differences and big numbers of difference-exceeding-tolerance cases in most of the tests. If the `Manual` if correct, then we'll see small maximum encountered difference in most of the tests. Still, due to the aforementioned truncation error, there may be some cases of large difference even when the `Manual` module produces correct Jacobians. These cases should be checked independently, e.g. by comparing `Manual`'s Jacobians to those produced by some other AD tool.
```
awf: Agreed -- ultimately we will not find *all* bugs in Manual this way, but we expect to find a useful number.

src/cpp/modules/finite/finite.h
```diff
@@ -117,7 +117,7 @@ class FiniteDifferencesEngine
             tmp_b = originalInput - delta;
             T dx = delta * 2;
             tmp_f = tmp_b + dx;
-            // adjusting dx so that (tmp_b + dx) - tmp_b == delta
+            // adjusting dx so that (tmp_b + dx) - tmp_b == 2 * delta
```
awf:
```
// adjusting dx so that (tmp_b + dx) - tmp_b == dx
```
In order to be absolutely clear it's about associativity.   As now, the reader needs to find the definition of dx, and to reassure themselves that dx == delta*2 exactly.

msdmkats: Agreed.

docs/Architecture.md
msdmkats: Yes, consecutiveness is indeed not necessary.

docs/JacobianCheck.md
msdmkats: No I didn't. I picked `1e-6`, because existing python tools produced only 6 decimal digits. We'll probably change that, so we might as well decrease the default tolerance. I'll run eigen tests against tolerance of `1e-8` overnight and see if there're any problems.

I can say right away, that finite differences are less precise, though. But that's the reason we can have separate tolerances for different tools.

awf: Yes, finite will always be less precise, but assuming Manual is correct, then |Tool - Manual| should be much better than finite differences.

msdmkats: Agreed. So eigen and eigenvector are fine with tolerance of `1e-8`. Let that be default. I set the tolerance for finite to `1e-5` so that it wouldn't produce too many errors.

PR #86, opened 2019-09-16
 Add Azure DevOps build definition for all the tools that we know work 

tomjaguarpaw:
Add Azure DevOps build definition for all the tools that we know work

### DIFF
--- a/ADBench/azure-devops-ci.ps1
+++ b/ADBench/azure-devops-ci.ps1
@@ -1,4 +1,11 @@
-python -m pip install --user matplotlib plotly
+# NB Julia is astonishingly slow to install.  10 mins for Julia itself
+# and the two needed packages.  Eventually we should try to improve
+# that, but for now we will live with it.
+choco install --limit-output julia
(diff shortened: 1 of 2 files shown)


tomjaguarpaw:
It only runs of four points of GMM at the moment, in the interests of finishing quickly.  Future work will involve

* Running on more points of GMM
* Running on new tools as they are added
* Running in an automated fashion and posting graphs to somewhere public
* Running for all new PRs

tomjaguarpaw:
> it would be really nice to run the unit tests that we have. It can be done by running ctest.

I can run the tests after running `ninja` in https://github.com/awf/ADBench/blob/30beba4c9d21894f6c4372e99098b4780952e31d/ADBench/azure-devops-ci.ps1

Can you give me the exact command line I should use?

msdmkats:
`ctest --output-on-failure`

That, probably, wouldn't fail the build on test failure by itself, though. For the future plans, I'd suggest having separate build steps for installing dependencies, building, running unit tests, and running performance tests, with every step requiring the previous one to succeed.

PR #87, opened 2019-09-16
Toelli/graph

tomjaguarpaw:
Ensure that lines close to each other don't have similar colours

(and other small graph tweaks)


awf:
Sounds great -- can you paste a before-and-after?

tomjaguarpaw:
Before is tricky because I'm in the middle of a new run.  Here's after:

image

awf:
Looks good.

PR #93, opened 2019-09-20
Precise C++ Output + C++ modules tests

novikov-alexander:
- C++ runner outputs precise values for objective and jacobian
- C++ module tests became harder due to more precise values received from fixed output
- An absolute error for C++ module tests is parametrized now because some algorithms give more precise results than another ones.


src/cpp/runner/OutputSave.cpp
```diff
@@ -17,20 +17,22 @@ std::string jacobian_file_name(const std::string& output_prefix, const std::stri
...
     out << std::scientific << objective_time << std::endl << derivative_time;
     out.close();
 }
 
 void save_value_to_file(const std::string& filepath, const double& value)
 {
-    std::ofstream out(filepath);
+    precise_ofstream<std::remove_reference_t<decltype(value)>> out(filepath);
+
     out << std::scientific << value;
```
msdmkats: Do we need `out << std::scientific` here? It seems to be already done in the constructor of `precise_ofstream<T>`.

test/cpp/modules/common/GmmTests.cpp
```diff
@@ -22,6 +22,8 @@ TEST_P(GmmModuleTest, Load)
     ASSERT_NE(module, nullptr);
 }
 
+const double epsilon = 1e-08;
```
msdmkats: Did you mean to remove this const after introducing a test parameter for tolerance?

test/cpp/modules/common/GmmTests.cpp
```diff
@@ -35,7 +37,7 @@ TEST_P(GmmModuleTest, ObjectiveCalculationCorrectness)
     module->calculate_objective(1);
 
     auto output = module->output();
-    EXPECT_NEAR(8.0738, output.objective, 0.00001);
+    EXPECT_NEAR(8.07380408004975791e+00, output.objective, 0.00001);
```
msdmkats: Probably, you wanted to remove the hard-coded tolerance from here as well.

msdmkats:
One more thing: change in the test structure (the new parameter for allowed tolerance) needs to be reflected in the doc.

test/cpp/modules/common/LstmTests.cpp
```diff
@@ -82,7 +82,7 @@ TEST_P(LstmModuleTest, JacobianCalculationCorrectness)
         0.000921, 0.000184, 0.000585, -0.000912, -0.003719, 0.000524, -0.000009,
         -0.000009, -0.000062, -0.000260, 0.002886, 0.001569, 0.001929, 0.000516,
         0.002204, 0.001783, 0.000347, -0.003243, -0.010870, 0.000814, -0.000509,
-        -0.000250, -0.000274, -0.000716, 0.002362, 0.000682, 0.002212, 0.005815, 
+        -0.000250, -0.000274, -0.000716, 0.002362, 0.000682, 0.002212, 0.005815,
         0.001197, 0.000226, 0.011221, -0.002159, -0.015211, 0.002407, -0.000016,
         -0.000016, -0.000099, -0.000323, 0.000009, 0.000000, 0.000000, 0.000000,
         0.000000, 0.000000, 0.000000, -0.000174, -0.000322, 0.000008, -0.000030,
```
mikhailnikolaev: I think you should remove hardcoded epsilon in this test class (lines 36 and 98)

PR #97, opened 2019-09-24
Use forward slashes in printed names instead of backslashes

tomjaguarpaw:
Patch implemented by @mikhailnikolaev


tomjaguarpaw:
@mikhailnikolaev Since you wrote this I guess you can just go ahead and merge it!

PR #107, opened 2019-09-28
Toelli/graph

tomjaguarpaw:
Simplified version of https://github.com/awf/ADBench/pull/96

### DIFF
--- a/ADBench/plot_graphs.py
+++ b/ADBench/plot_graphs.py
@@ -226,15 +226,18 @@ for (figure_idx, (graph, function_type)) in enumerate(all_graphs, start=1):
 
     # Plot results
     for ((color, marker), (tool, n_vals, t_vals, violations)) in lines:
-        # Checking neighbours by shifting t_vals requires that
-        # it is in the order of monotonic n_vals
-        assert n_vals == sorted(n_vals)
-        together = list(zip(
-            n_vals,
-            t_vals,
-            [True] + [t_val == float("inf") for t_val in t_vals],
-            [t_val == float("inf") for t_val in t_vals[1:]] + [True],
-            violations))
+        def vals_with_neighbours_and_violation():
+            # Checking neighbours by shifting t_vals requires that
+            # it is in the order of monotonic n_vals
+            assert n_vals == sorted(n_vals)
+            return zip(
+                n_vals,
+                t_vals,
+                [True] + [t_val == float("inf") for t_val in t_vals],
+                [t_val == float("inf") for t_val in t_vals[1:]] + [True],
+                violations)
+
+        together = list(vals_with_neighbours_and_violation())
 
         all_terminated = all(t_val == float("inf") for t_val in t_vals)
 
--- a/ADBench/plot_graphs.py
+++ b/ADBench/plot_graphs.py
@@ -266,7 +266,7 @@ for (figure_idx, (graph, function_type)) in enumerate(all_graphs, start=1):
         additional.append(([n_val for (n_val, _) in additionals],
                            [t_val for (_, t_val) in additionals],
                            color))
-                
+
         violation_x += [n_val for (n_val, _, _, _, violation)
                         in together if violation]
         violation_y += [t_val for (_, t_val, _, _, violation)
--- a/ADBench/plot_graphs.py
+++ b/ADBench/plot_graphs.py
@@ -236,8 +236,7 @@ for (figure_idx, (graph, function_type)) in enumerate(all_graphs, start=1):
             [t_val == float("inf") for t_val in t_vals[1:]] + [True],
             violations))
 
-        all_terminated = all(t_val == float("inf")
-                             for t_val in t_vals)
+        all_terminated = all(t_val == float("inf") for t_val in t_vals)
 
         label = utils.format_tool(tool)
         # Append label in legend if all point values are infinite
--- a/ADBench/plot_graphs.py
+++ b/ADBench/plot_graphs.py
@@ -264,7 +264,7 @@ for (figure_idx, (graph, function_type)) in enumerate(all_graphs, start=1):
                        and missing_right
                        and violation]
 
-        # addint coordinates of additional markers
+        # adding coordinates of additional markers
         additional.append(([n_val for (n_val, _) in additionals],
                            [t_val for (_, t_val) in additionals],
                            color))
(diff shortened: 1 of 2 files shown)


tomjaguarpaw:
@awf This contains the changes you requested on #96

tomjaguarpaw:
@mikhailnikolaev This contains a number of changes to your original which I consider to be simplifications.  Please take a look at the commits and tell me what you think.

I suspect it's best to squash merge this.

ADBench/plot_graphs.py
```diff
@@ -99,7 +130,7 @@ def read_vals(objective, graph_files, tool):
     t_objective_vals = list(map(lambda t: t[1][0], info_sorted))
     t_jacobian_vals = list(map(lambda t: t[1][1], info_sorted))
 
-    return (n_vals, t_objective_vals, t_jacobian_vals)
+    return (n_vals, t_objective_vals, t_jacobian_vals, violation_vals)
```
mikhailnikolaev: We need to sort violation_vals as I proposed in commit https://github.com/awf/ADBench/pull/96/commits/cecc228fa21929955b58a921c1d7799627e9b87e

mikhailnikolaev: We need to sort violation_vals as I proposed in the commit https://github.com/awf/ADBench/commit/cecc228fa21929955b58a921c1d7799627e9b87e

ADBench/plot_graphs.py
```diff
@@ -188,25 +219,78 @@ def sorting_key_fun(v):
...
-        failed_x.append(handle.get_xdata()[last_ind])
-        failed_y.append(handle.get_ydata()[last_ind])
-        #handles += pyplot.plot(handle.get_xdata()[last_ind], handle.get_ydata()[last_ind], marker="o", color=(0, 0, 0), linestyle="None", label="Crashed/Terminated")
-    if len(failed_x) > 0:
-        handles += tuple(pyplot.plot(failed_x, failed_y, marker="o", color="k", linestyle="None", label=TERMINATED_LABEL))
-        labels += (TERMINATED_LABEL,)
+    violation_x, violation_y = [], []
+    was_violation = False
+    additional = []
+    violation_handle = None
```
mikhailnikolaev: As I understand, this is useless in code after your changes

ADBench/plot_graphs.py
```diff
@@ -188,25 +219,78 @@ def sorting_key_fun(v):
...
+
+    # Plot results
+    for ((color, marker), (tool, n_vals, t_vals, violations)) in lines:
+        def vals_with_neighbours_and_violation():
+            # Checking neighbours by shifting t_vals requires that
+            # it is in the order of monotonic n_vals
+            assert n_vals == sorted(n_vals)
+            return zip(
```
mikhailnikolaev: Looks very nice! I really like this code (much more than my version of it)

ADBench/plot_graphs.py
```diff
@@ -188,25 +219,78 @@ def sorting_key_fun(v):
...
+        violation_x += [n_val for (n_val, _, _, _, violation)
+                        in together if violation]
+        violation_y += [t_val for (_, t_val, _, _, violation)
+                        in together if violation]
+        was_violation = \
+            was_violation \
+            or any(violation for (_, _, _, _, violation) in together)
+
+
+    handles += pyplot.plot(
```
mikhailnikolaev: I think, handle for violation points need to be added only if there was any violation, so, in my opinion, it should be moved to "if" body below

ADBench/plot_graphs.py
```diff
@@ -188,25 +219,78 @@ def sorting_key_fun(v):
...
+        together = list(vals_with_neighbours_and_violation())
+        additionals = [(n_val, t_val)
+                       for (n_val, t_val, missing_left, missing_right, violation)
+                       in together
+                       if t_val != float("inf")
+                       and missing_left
+                       and missing_right
+                       and violation]
+
+        # addint coordinates of additional markers
```
mikhailnikolaev: Misprint

ADBench/plot_graphs.py
tomjaguarpaw: Yes, you're right.  Thanks!

ADBench/plot_graphs.py
tomjaguarpaw: We could have it inside the `if` or outside.  The behaviour will be the same.  The question is, what leads to simpler code.  I think it's simpler to have it outside.  Then we don't have to think about whether it's run or not.  It is always run and always does the right thing, even if there are no violations.  What do you think?

ADBench/plot_graphs.py
tomjaguarpaw: Well spotted!  Thank you.

tomjaguarpaw: Done

ADBench/plot_graphs.py
tomjaguarpaw: Done

ADBench/plot_graphs.py
tomjaguarpaw: Done

ADBench/plot_graphs.py
tomjaguarpaw: Thanks!  Hopefully you can see how it is possible to use Python iterators to avoid explicit indexing.  Explicit indexing can get very messy and can be hard to understand.

ADBench/plot_graphs.py
mikhailnikolaev: This make sense for plotly, because when it will convert figure it will look through handles and will make legend by them, not by labels. So, here the example of your code result in plotly:
image
AS you can see, there is not any violations but violation handle is present

tomjaguarpaw: Ah that's interesting.  Thank you.  I will fix that.

tomjaguarpaw: Done

ADBench/plot_graphs.py
```diff
@@ -23,8 +23,10 @@
```
mikhailnikolaev: I have found something interesting related to lines 21-22. If user runs this script to get a reference (`python plot_graphs.py --help`, for example) then warning message about `--show` flag enabling will be printed. That is because in this case `do_plotly` and `do_save` are false, so, `do_show` will be true. I think 21-22 lines shoul be mooved in place after reference printing (to line 196). What do you think?

tomjaguarpaw: Good point.  This PR helps a bit https://github.com/awf/ADBench/pull/108

But we should subsequently refactor the script so no executable code runs at the top level.

PR #113, opened 2019-10-02
Dotnet runner and DIffSharp module

mikhailnikolaev:
Work that has been done (by @novikov-alexander):
1. Dotnet runner is implemented.
2. **BA** objective differentiation using _DiffSharp_ is implemented.
3. Unit tests for runner and for **BA** processing are added.

# Grpahs
**Note:** the new version of DiffSharp module is called _DiffSharpModule_ in these graphs.
**Note:** the old version of DiffSharp is marked as incorrect because it produces output files in a wrong format.
BA  Jacobian  - Release Graph
BA  Jacobian ÷ objective  - Release Graph
BA  Objective  - Release Graph
BA  Objective ÷ manual  - Release Graph


tomjaguarpaw:
DiffSharpModule also looks generally faster than DiffSharp.  Is that right?  If so, that's great!

awf:
Is this because of BASparseMatrix.cs?  Interesting, and valid, as the AD itself doesn't need to pass through it.

src/dotnet/runner/Data/BAData.cs
```diff
@@ -0,0 +1,105 @@
...
+
+namespace DotnetRunner.Data
+{
+
+    // rows is nrows+1 vector containing
+    // indices to cols and vals. 
+    // rows[i] ... rows[i+1]-1 are elements of i-th row
+    // i.e. cols[row[i]] is the column of the first
+    // element in the row. Similarly for values.
+    public class BASparseMatrix
```
awf: Ideally we would do this in F#, or use a generic SparseMatrix class in an F# numerics library.
Do you know of one?

src/dotnet/modules/DiffSharpModule/ba.fs
```diff
@@ -0,0 +1,206 @@
...
+let create_sparse_J n m p (obs:_[][]) (reproj_err_d:_[,][]) (w_err_d:_[]) =
+    let nrows = 2 * p + p;
+    let ncols = N_CAM_PARAMS*n + 3 * m + p;
+
+    let to1DArray (arr: 'T [,]) =
+        seq { for y in [0..(Array2D.length2 arr) - 1] do 
+                  for x in [0..(Array2D.length1 arr) - 1] do 
+                      yield arr.[x, y] } |> Seq.toArray
+
+    let J = new BASparseMatrix(n, m, p);
```
awf: See below - can we not do this in F#?

src/dotnet/modules/DiffSharpModule/ba.fs
```diff
@@ -0,0 +1,204 @@
...
+
+let rodrigues_rotate_point (rot:DV) (x:DV) =
+    let sqtheta = DV.l2normSq rot
+    if sqtheta <> D 0. then
+        let theta = sqrt sqtheta
+        let costheta = cos theta
+        let sintheta = sin theta
+        let theta_inv = D 1. / theta
+
+        let w = rot * theta_inv
```
mikhailnikolaev: I don't see where theta_inv is used except the line 32. So, maybe it will be more efficient to merge these lines

msdmkats: Not sure about efficiency, but I see no reason for `theta_inv`'s existence as well. So merged.

PR #114, opened 2019-10-04
Old Tapenade and Eigen code removing

mikhailnikolaev:
Old code from directories *tools/Tapenade* and *tools/cpp-common* is removed due to uselessness


msdmkats:
Did you check, if Adept, ADOLC, and Ceres can be built without the files you removed from cpp-common?

mikhailnikolaev:
I've found one place where the old code was used. Now it isn't using anywhere

PR #116, opened 2019-10-08
CPP common test improvement

mikhailnikolaev:
1. BA common test checking jacobian is extended (now it asserts all the values of BA sparse jacobian)
2. GMM test checking is refactored in a vector form
3. _manualTests_ are renamed to _commonTests_
4. Test values are flattened.


tomjaguarpaw:
This looks like a good improvment.

PR #121, opened 2019-10-09
Zygote BA uses reverse mode.

msdmkats:
BA  Jacobian  - Release Graph

Zygote is zygote in reverse mode
ZygoteF uses Zygote.forward_jacobian

### DIFF
--- a/src/julia/modules/Zygote/ZygoteBA.jl
+++ b/src/julia/modules/Zygote/ZygoteBA.jl
@@ -85,7 +85,9 @@ function compute_ba_J(cams, X, w, obs, feats)
         compute_reproj_err_d_i = x -> compute_reproj_err_d(x, feats[:, i])
         camIdx =  obs[1, i]
         ptIdx = obs[2, i]
-        _, J = Zygote.forward_jacobian(compute_reproj_err_d_i, pack(cams[:, camIdx], X[:, ptIdx], w[i]))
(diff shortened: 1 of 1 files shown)


src/julia/modules/Zygote/ZygoteBA.jl
```diff
@@ -85,7 +85,9 @@ function compute_ba_J(cams, X, w, obs, feats)
         compute_reproj_err_d_i = x -> compute_reproj_err_d(x, feats[:, i])
         camIdx =  obs[1, i]
         ptIdx = obs[2, i]
-        _, J = Zygote.forward_jacobian(compute_reproj_err_d_i, pack(cams[:, camIdx], X[:, ptIdx], w[i]))
+        y, back = Zygote.forward(compute_reproj_err_d_i, pack(cams[:, camIdx], X[:, ptIdx], w[i]))
```
awf: Is this reverse mode?

msdmkats: Yes. This call produces a function `back` that computes the derivative for a given direction at the point specified int the `forward` call.
https://fluxml.ai/Zygote.jl/stable/adjoints/

awf: Thanks, sorry for the noise.

tomjaguarpaw: Yes, the terminology is rather unfortunate.  There's "forward mode AD" and "reverse mode AD forward pass".

PR #123, opened 2019-10-09
Changes needed to compile with g++

tomjaguarpaw:
These are needed for g++ and seem to be compatible with Visual C++ too, at least the build succeds.


src/cpp/shared/utils.h
```diff
@@ -94,10 +95,10 @@ class write_J_stream : public precise_ofstream<T>
 {
 public:
     write_J_stream(std::string fn, size_t rows, size_t cols)
-        :precise_ofstream(fn)
+        :precise_ofstream<T>(fn)
     {
         std::cout << "Writing to " << fn << std::endl;
-        if (!good()) {
+        if (this->good()) {
```
mikhailnikolaev: So, why this condition was changed to the opposite one?

tomjaguarpaw: Ah, well spotted!  I'm glad you are giving this code review.

I will have to check why I did not notice any errors from this though.

tomjaguarpaw: Thank you very much for catching my error.  I think the reason that I did not notice this is that it is only used by the C++ tools that have not yet been migrated to your new runner.

mikhailnikolaev: You're welcome

Issue #118, opened 2019-10-08
Consistent naming


awf:
We should name tools with their tool name and the language/platform, e.g. Julia-Zygote, Python-Autograd.  C++-Finite, C++-Manual, C++-ManualEigen

awf:
At least on the graphs, doesn't need to be root-and-branch.

mikhailnikolaev:
What about this name style: `Tool (language/platform)`? For example `Manual (C++)`

awf:
I was thinking language first, as that will tend to match sorting order.

mikhailnikolaev:
> I was thinking language first, as that will tend to match sorting order.

So, what about `(C++) Manual` ? I propose bracket because, for example, string `C++-Manual` looks rather strange

awf:
I've tried a few, I thinl comma is good:
```
C++, Manual
Julia, Zygote
Python, Autograd
```

mikhailnikolaev:
So, could we close this issue due to PR #133?

awf:
Yes, thanks,

Issue #109, opened 2019-09-30
Matters arising from plot_graphs.py


tomjaguarpaw:
Hi @mikhailnikolaev, thanks very much for all your work on the new graph drawing code.  The new graphs look a lot better than the old ones, contain more information, and the code is more coherent too.  There are a few matter that arise from that that we should consider:

1. Your original PR contained a lot of whitespace changes.  Personally, I'm happy with the formatting of the as it currently is.  On the other hand I'm open to discussion and I'm not particularly opposed to whitespace changes.  If we're going to do that we should decide on the format (and perhaps choose an autoformatter) and do the whole file in one whitespace-only commit.  Whitespace changes that occur at the same time as refactorings or functional changes make diffs hard to read.
       a. I am totally fine with wrapping long lines though
2. A lot of the processing occurs at the module top level.  This is pretty bad form.  I think we have two options:
    1. If we're not going to make any more changes to that file then I think it's fine to leave it as it is.  It works.  Let's not try to fix what works.
    2. If we make more changes to this file then first we should pull all executable code into functions and have a `main()` function that is the entry point of the script.  Then new functional changes can be made.

What do you think?

mikhailnikolaev:
Hello, @tomjaguarpaw. I think right now we have others topics to do, so, if current code works, then maybe we can do some refactoring later. (But I think the idea about `main` is absolute right)

tomjaguarpaw:
OK, I'll give awf a chance to look at the outstanding PR and then merge it. Let's come back to these things if we touch the file again in the future.

> By the way, thank you for your help

You're welcome!  I hope it was useful for your understanding.

tomjaguarpaw:
@mikhailnikolaev says:

> I thought when all top-level code was in the same place of the script, code structuring could be simplier. I proposed that such a structuring could be done in the future PR. More over I absolutely don't like when the top level code is not in `if __name__ == "__main__":` block

Yes, I agree with your intentions but I think we need to be careful with `plot_graphs.py` because it is already complicated and fragile.  I suggest that instead of moving everything at the top-level into an `if __name__ == "__main__"` block we move specific segments of the top-level code into functions.  Once we've done that in a way that makes sense then we'll think about how to get rid of the rest of the code from the top-level.

tomjaguarpaw:
I think we have sorted this nicely.  Almost everything lives outside the top level.  There's a tiny bit still left to do but no need to keep this issue open.

PR #127, opened 2019-10-11
Graph plotting improvement

mikhailnikolaev:
1. Each tool has the same color on each graph generated in the one run of the graph plotting script. That means that their colors can differ from run to run if, for example, a new tool has been added between runs.

2. Vertical line over each X point is added for better understanding of graphs.
**Note**: such lines disappear in plotly graphs by some reason.

# Examples
BA  Jacobian  - Release Graph
GMM (1k)  Jacobian ÷ objective  - Release Graph
HAND (Complicated, Big)  Objective  - Release Graph

# Plotly example
image

### DIFF
--- a/ADBench/plot_graphs.py
+++ b/ADBench/plot_graphs.py
@@ -285,8 +285,11 @@ CMD arguments:
                     [t_val == float("inf") for t_val in t_vals[1:]] + [True],
                     violations)
 
-            color, marker = get_style(tool, all_tool_names)
+            # do nothing in case when there is not any values for this tool
+            if len(t_vals) == 0:
+                continue
 
+            color, marker = get_style(tool, all_tool_names)
             label = utils.format_tool(tool)
             all_terminated = all(t_val == float("inf") for t_val in t_vals)
             # Append label in legend if all point values are infinite
--- a/ADBench/plot_graphs.py
+++ b/ADBench/plot_graphs.py
@@ -175,6 +175,17 @@ def get_style(tool, tool_names):
 
 
 
+def  draw_vertical_lines(sorted_vals_by_tools):
+    '''Adds vertical lines to the figure for clarifying results.'''
+
+    n_vals = [ n_val for _, n_val, _, _ in sorted_vals_by_tool ]
+    max_n_val = max(n_vals, key = lambda n_val: len(n_val))
+
+    for n in max_n_val:
+        pyplot.axvline(n, ls = '--', color = "grey", zorder = 0.0, lw = 0.5)
+
+
+
 if __name__ == "__main__":
     # Script arguments
     do_save = "--save" in sys.argv
@@ -339,6 +350,8 @@ CMD arguments:
         pyplot.ylabel(f"Running time (s) for [{function_type.capitalize()}]")
         pyplot.xscale("log")
         pyplot.yscale("log")
+
+        draw_vertical_lines(sorted_vals_by_tool)
 
         # Export to plotly (if selected)
         if do_plotly:
--- a/ADBench/plot_graphs.py
+++ b/ADBench/plot_graphs.py
@@ -182,7 +182,7 @@ def  draw_vertical_lines(sorted_vals_by_tools):
     max_n_val = max(n_vals, key = lambda n_val: len(n_val))
 
     for n in max_n_val:
-        pyplot.axvline(n, ls = '--', color = "grey", zorder = 0.0, lw = 0.5)
+        pyplot.axvline(n, ls = '-', color = "lightgrey", zorder = 0.0, lw = 0.5)
 
 
 
--- a/ADBench/plot_graphs.py
+++ b/ADBench/plot_graphs.py
@@ -35,6 +35,7 @@ static_out_dir = os.path.join(out_dir, static_out_dir_rel)
 plotly_out_dir = os.path.join(out_dir, plotly_out_dir_rel)
 
 
+
 def graph_data(build_type, objective, maybe_test_size):
     '''Creates graph name and graph saving location.'''
 
@@ -49,7 +50,11 @@ def graph_data(build_type, objective, maybe_test_size):
 
     return (graph_name, graph_save_location)
 
+
+
 has_manual = lambda tool: tool.lower() in ["manual", "manual_eigen"]
+
+
 
 def tool_names(all_files):
     '''Returns a set of tool names from all calculated files.'''
@@ -65,6 +70,8 @@ def tool_names(all_files):
     print(tool_names_)
 
     return tool_names_
+
+
 
 def read_vals(objective, graph_files, tool):
     '''Extracts data for files of the specified tool.'''
@@ -116,6 +123,8 @@ def read_vals(objective, graph_files, tool):
 
     return (n_vals, t_objective_vals, t_jacobian_vals, violation_vals)
 
+
+
 def vals_by_tool(objective, graph_files, tool_names):
     '''Classifies file values by tools'''
 
@@ -150,6 +159,8 @@ def vals_by_tool(objective, graph_files, tool_names):
             raise Exception(f"Unknown function type {function_type}")
 
         yield (tool, n_vals, t_vals, violation)
+
+
 
 def get_style(tool, tool_names):
     '''Returns the style of the given tool from all tool list.'''
@@ -216,6 +227,8 @@ CMD arguments:
     all_graphs = [(path, function_type) for function_type in function_types for path in all_graphs]
     all_graph_dict = {}
 
+
+
     # Loop through each of graphs to be created
     for (figure_idx, (graph, function_type)) in enumerate(all_graphs, start=1):
         build_type = graph[0]
@@ -236,6 +249,7 @@ CMD arguments:
                 return sum(y_list) / len(y_list)
             else:
                 return 1e9
+
         sorted_vals_by_tool = sorted(vals_by_tool(objective, graph_files, all_tool_names),
                                     key=sorting_key_fun,
                                     reverse=True)
--- a/ADBench/plot_graphs.py
+++ b/ADBench/plot_graphs.py
@@ -51,10 +51,13 @@ def graph_data(build_type, objective, maybe_test_size):
 
 has_manual = lambda tool: tool.lower() in ["manual", "manual_eigen"]
 
-def tool_names(graph_files):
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
That looks good.  @awf, is this what you were looking for?

@mikhailnikolaev Thanks for doing this.  I will give it a thorough review as soon as I can, probably Monday.

ADBench/plot_graphs.py
```diff
@@ -172,8 +162,40 @@ def div_lists(alist, blist):
 
 
 
-if do_help:
-    ref_msg = f'''
+def get_style(tool, tool_names):
+    '''Returns the style of the given tool from all tool list.'''
+
+    idx = tool_names.index(tool)
```
awf: But why not just hash the name?  This means new tools will change the colours.  And we had to sort etc.  The chance of collision is exactly the same.

ADBench/plot_graphs.py
```diff
@@ -172,8 +162,40 @@ def div_lists(alist, blist):
...
+
+
+def  draw_vertical_lines(sorted_vals_by_tools):
+    '''Adds vertical lines to the figure for clarifying results.'''
+
+    n_vals = [ n_val for _, n_val, _, _ in sorted_vals_by_tool ]
+    max_n_val = max(n_vals, key = lambda n_val: len(n_val))
+
+    for n in max_n_val:
+        pyplot.axvline(n, ls = '--', color = "grey", zorder = 0.0, lw = 0.5)
```
awf: Can we use non dashes and light grey please.

awf:
Great, thanks.  A couple of comments.

ADBench/plot_graphs.py
mikhailnikolaev: But this way we will never have collisions, except the count of tools is greater than count of styles, which can always be expanded if it necessary. Honestly, I think for this simple case hashing is too powerful solution in differ with simple tool name sorting. But it is only my opinion

ADBench/plot_graphs.py
mikhailnikolaev: Of course, I will change it as soon as possible

mikhailnikolaev: Changed. Is this better?

ADBench/plot_graphs.py
awf: Oh yes, sorry.  Agreed - chance of collision is indeed high.

ADBench/plot_graphs.py
awf: Great thanks.

awf:
Folks, can we generate graphs with Julia, TF, PyTorch, and the C++ models by Monday PM?

ADBench/plot_graphs.py
awf: So remove the sorting, and just use a fixed mapping toolname -> style as initially discussed.

ADBench/plot_graphs.py
```diff
@@ -192,176 +214,201 @@ def div_lists(alist, blist):
...
+            color, marker = get_style(tool, all_tool_names)
+
+            label = utils.format_tool(tool)
+            all_terminated = all(t_val == float("inf") for t_val in t_vals)
+            # Append label in legend if all point values are infinite
+            if all_terminated:
+                label += ALL_TERMINATED_SUFFIX
+
+            labels.append(label)
+            handles += pyplot.plot(
```
msdmkats: I think, it should be enough to not append the label if `t_vals` is empty.

ADBench/plot_graphs.py
msdmkats: Maybe, just fix the legend? https://github.com/awf/ADBench/pull/127#discussion_r334376092

tomjaguarpaw:
Some of these changes are uncontroversial refactorings or tweakings.  Could they please be submitted as a separate PR which we can get through quickly?  (Not the `__main__` change though please.  That's bigger and I want to think about how we structure this script in general.)

tomjaguarpaw:
I pulled out the tweaks into a separate PR: https://github.com/awf/ADBench/pull/129

I suggest we merge that first and then we can continue with the changes affecting the output.

ADBench/plot_graphs.py
tomjaguarpaw: I suggest that we drop the change to avoid recalculating tool names.  Sure it's a little bit faster, but my main concern with this file is how easy it is to read.  It's already far more complicated than necessary and I think we should avoid changing the logic before we find a way of making it considerably simpler.

mikhailnikolaev: > I suggest that we drop the change to avoid recalculating tool names. Sure it's a little bit faster, but my main concern with this file is how easy it is to read. It's already far more complicated than necessary and I think we should avoid changing the logic before we find a way of making it considerably simpler.

What did you mean by `avoid recalculating tool names`? Now they are calculated only once at the script beginning

mikhailnikolaev:
Now nonexistent tools just are not added as Dmitry requested

mikhailnikolaev:
> Would prefer that the top-level code not just be moved under an `if` like this. We will discuss further.

I absolutely agree with this. I've made such a refactoring only as the first step of future code structuring, I thought when all top-level code was in the same place of the script, code structuring could be simplier. I proposed that such a structuring could be done in the future PR. More over I absolutely don't like when the top level code is not in `if __name__ == "__main__:` block

ADBench/plot_graphs.py
mikhailnikolaev: > So remove the sorting, and just use a fixed mapping toolname -> style as initially discussed.

I think after my the last commit it is still unnecessary

awf:
I really do want the static mapping name to colours.   What is the downside?

mikhailnikolaev:
> I really do want the static mapping name to colours. What is the downside?

I think the main downside is a possibility of collisions itself. But if you really do want the static mapping I can do this. I think that that way this PR should be closed, becuase the commits about vertical lines and some litle refactoring are moved to another PR (https://github.com/awf/ADBench/pull/129) by Tom and the rest commits are all about nonstatic mapping or about top-level code refactoring that Tom is not happy about. What do you think?

awf:
> I really do want the static mapping name to colours. What is the downside?
> 
> I think the main downside is a possibility of collisions itself. But if you really do want the static mapping I can do this. I think that way this PR should be closed, becuase the commits about vertical lines and some litle refactoring are moved to another PR (#129) by Tom and the rest commits are all about nonstatic mapping or about top-level code refactoring that Tom is not happy about. What do you think?

What I mean is a global mapping from tool to style, initialized from a list, with the ordering on the legend either as in the list, or based on the value of the leftmost datapoint.
```
tool_styles = {
   'Autograd', ('r', '--'),  # Tool name, color, style, optionally tool renaming
   'Zygote', ('g', '-', 'Julia Zygote'),
   'Manual',  ('k', '-'. 'C++Manual')
};
```

tomjaguarpaw:
I think the below summarises the status of all the commits in this PR.  There is only one outstanding which should be addressed as far as possible according to @awf's specification above and submitted in a new PR.  Therefore I am closing this PR.

* [x]  All main logic is moved to main section  https://github.com/awf/ADBench/pull/127/commits/ba5f423f1a76eb6146c7a798a71292afd5e2fa37
    Discuss at https://github.com/awf/ADBench/issues/109
* [x]  Message about output directory is shown only if necessary https://github.com/awf/ADBench/pull/127/commits/d993d43ca23ccc19b11935c5b825542d5535ae6a
    Merged
* [x]  Info and warning output is moved to the main beginning 
    Discuss at https://github.com/awf/ADBench/issues/109
* [x] `--show` warning info is splitted to two lines  https://github.com/awf/ADBench/pull/127/commits/30f371e917937652e20aedc43637561bb5f0bbec
    Merged
* [x]  Tool name once calculating https://github.com/awf/ADBench/pull/127/commits/7b8e67cac8b5a8f976bf5a7cd7d21c03ea818598
    Discuss at https://github.com/awf/ADBench/issues/109
* [ ]  Consistent styles 
    TDB
* [x]  Info output order is changed  https://github.com/awf/ADBench/pull/127/commits/90c118ad26768c0c1ba59fd5b1aa36c45fdc9c1b
    Merged
* [x]  Separating line adding https://github.com/awf/ADBench/pull/127/commits/b97e72e9200a22fd94e3ce9fd74a992f5209b004
    Discuss at https://github.com/awf/ADBench/issues/109
* [x]     Vertical line adding  https://github.com/awf/ADBench/pull/127/commits/4302425b64d89bbe4c14281b72baacf156853379
    Merged
* [x]  Vertical line style changing  https://github.com/awf/ADBench/pull/127/commits/4710c94923d3f20c65694176b837e63472940ccc
    Merged
* [x] Not marking nonexistence results as crashed https://github.com/awf/ADBench/pull/127/commits/726847f0163a3dee7a2b1cfe6779c99f3802a0f4
    Discuss at https://github.com/awf/ADBench/issues/109 (not needed if we don't make other changes to `plot_graphs.py`

awf:
OK, but let's do consistent styles immediately.  It's fine to have changed our mind: I first thought hashing would work, Mikhail pointed out that it wouldn't, but then we went to one class of fix (sorting) when on reflection I believe another fix (static table) is better.

tomjaguarpaw:
> OK, but let's do consistent styles immediately

Yes, let's.  But we can't do it on this PR because it will no longer cleanly apply, so @mikhailnikolaev must port his fix to the current `master`.

mikhailnikolaev:
> > I really do want the static mapping name to colours. What is the downside?
> > I think the main downside is a possibility of collisions itself. But if you really do want the static mapping I can do this. I think that way this PR should be closed, becuase the commits about vertical lines and some litle refactoring are moved to another PR (#129) by Tom and the rest commits are all about nonstatic mapping or about top-level code refactoring that Tom is not happy about. What do you think?
> 
> What I mean is a global mapping from tool to style, initialized from a list, with the ordering on the legend either as in the list, or based on the value of the leftmost datapoint.
> 
> ```
> tool_styles = [
>    'Autograd', ('r', '--'),  # Tool name, color, style, optionally tool renaming
>    'Zygote', ('g', '-', 'Julia.Zygote'),
>    'Manual',  ('k', '-', 'C++.Manual')
> ]
> ```

I like this. Moreover it will solves the issue https://github.com/awf/ADBench/issues/118. I will do this as soon as possible.

PR #132, opened 2019-10-15
Plot graph refactor

tomjaguarpaw:
Lots of refactorings to move executable code out of the top level.  Almost everything is now done in functions.


tomjaguarpaw:
Thanks!

> Am I understand correctly, that now refactoring is done?

If you're asking because you want to go ahead and make your changes without worrying about the file changing underneath you then let's say yes, it's done and you can go ahead!

mikhailnikolaev:
> Thanks!
> 
> > Am I understand correctly, that now refactoring is done?
> 
> If you're asking because you want to go ahead and make your changes without worrying about the file changing underneath you then let's say yes, it's done and you can go ahead!

Perfect!

PR #133, opened 2019-10-15
Style mapping

mikhailnikolaev:
Styles are hard linked to tools by map. For each new tool respected style must be added to the map. If tool style is not specified, default style will be used. Map has the following form:
```Python
all_styles = {
    <tool name>: (<color>, <marker> [,<optional displaying name of the tool>]),
    ...
}
```
Using the same marks for the tools of the same language/platform is suggested.

# Plot examples
## Static
BA  Jacobian  - Release Graph
GMM (1k)  Jacobian ÷ objective  - Release Graph
HAND (Complicated, Big)  Objective  - Release Graph

## Plotly
image

### DIFF
--- a/ADBench/plot_graphs.py
+++ b/ADBench/plot_graphs.py
@@ -45,7 +45,7 @@ all_styles = {
 
     # Python tools
     "PyTorch": ("g", "s", "Python, PyTorch"),
-    "Tensorflow": ("r", "s", "Python, Tensorflow (eager execution)"),
+    "Tensorflow": ("r", "s", "Python, Tensorflow"),
     "Autograd": ("c", "s", "Python, Autograd"),
 
     # Julia tools
--- a/ADBench/plot_graphs.py
+++ b/ADBench/plot_graphs.py
@@ -29,9 +29,28 @@ ALL_TERMINATED_SUFFIX = " (all crashed/terminated)"
 figure_size = (9, 6) if do_plotly else (12, 8)
 fig_dpi = 96
 save_dpi = 144
-colors = ["b", "g", "r", "c", "m", "y"]
-markers = ["*", "+", "s", "^"]
-all_styles = [(c, m) for m in markers for c in colors]
+
+all_styles = {
+    # C++ tools
+    "Finite": ("b", "*"),
+    "FiniteEigen": ("g", "*"),
+    "Manual": ("r", "*"),
+    "ManualEigen": ("c", "*"),
+    "ManualEigenVector": ("m", "*"),
+    "Tapenade": ("y", "*"),
+
+    # .Net tools
+    "DiffSharp": ("b", "o"),
+
+    # Python tools
+    "PyTorch": ("g", "s"),
+    "Tensorflow": ("r", "s"),
+    "Autograd": ("c", "s"),
+
+    # Julia tools
+    "Julia": ("m", "v"),
+    "Zygote": ("y", "v")
+}
 
 # Folders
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
At the moment all unknown tools are drawn in black with "x".  I do think we still need to cycle through a list of styles for the unknown tools.

tomjaguarpaw:
I suggest this be done in a similar way to the original opproach of zipping with a list of styles.  You can pass `sorted_vals_by_tool` to a generator function which looks up the tool in the styles mapping and returns the style (and name) if present, or if not returns the next element from (the original) `all_styles` list.

mikhailnikolaev:
> I suggest this be done in a similar way to the original approach of zipping with a list of styles. You can pass `sorted_vals_by_tool` to a generator function which looks up the tool in the styles mapping and returns the style (and name) if present, or if not returns the next element from (the original) `all_styles` list.

I've done what you proposed, except one thing. I think that default styles should have the unique feature, so, we can realized that they are default. I make a marker such a feature. All defaults have 'x' marker (and several different colors). Thus, someone, who will add a new tool, will understand that 'x' marker should not be used in this tool style

ADBench/plot_graphs.py
```diff
@@ -29,9 +29,29 @@
...
+default_styles = [ (color, "x") for color in "rgbcmyk" ]
+tool_styles = {
+    # C++ tools
+    "Finite": ("b", "*", "C++, Finite"),
+    "FiniteEigen": ("g", "*", "C++, Finite Eigen"),
+    "Manual": ("r", "*", "C++, Manual"),
+    "ManualEigen": ("c", "*", "C++, Manual Eigen"),
+    "ManualEigenVector": ("m", "*", "C++, Manual Eigen Vector"),
+    "Tapenade": ("y", "*", "C, Tapenade"),
+
```
awf: Suggest colour = black for all Manual, and then distinguish via marker.  Can you also add linestyle?
colour = Blue for all Finite

ADBench/plot_graphs.py
```diff
@@ -29,9 +29,28 @@
...
+    # .Net tools
+    "DiffSharp": ("b", "o", ".Net, DiffSharp"),
```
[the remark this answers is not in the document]
mikhailnikolaev: Done

ADBench/plot_graphs.py
mikhailnikolaev: Do we actually need different linestyles?

mikhailnikolaev:
Strange, but plotly can not convert some kinds of markers. E.g. wide pluses are converted to rounds, and tiny diamonds are converted to regular ones

tomjaguarpaw:
This is very nice!

tomjaguarpaw:
> I think that default styles should have the unique feature, so, we can realized that they are default. I make a marker such a feature.

I think this is a good idea.

tomjaguarpaw:
[As a suggestion (not required) you could minimise the diff by keeping the names `color_marker` and `label`, and the parentheses around `color, marker` iterator unpacking, but that is completely optional depending on your taste.]

mikhailnikolaev:
> [As a suggestion (not required) you could minimise the diff by keeping the names `color_marker` and `label`, and the parentheses around `color, marker` iterator unpacking, but that is completely optional depending on your taste.]

Sorry, I saw your suggestion after I had merged the PR

tomjaguarpaw:
That's no problem at all.

PR #134, opened 2019-10-17
Docker support

iliaeg:
Implemented:
- python auto requirements install on cmake config stage
- shell wrapper script for docker use (It might be used from linux either)
- dockerfile with .dockerignore


iliaeg:
Also I've written short guide. It is not ready yet, but may be useful for current use.

## Build docker image

1. Clone repository
2. Add nessasary AD tools
3. Run `docker build`  
   e.g. `docker build -t adb-docker .`

## Use docker image

1. Create folder for results
2. Run docker image with proper arguments  
   Use option `-r|--run` to run benchmark  
   e.g. `docker run -v /folder/for/results:/adb/tmp/ adb-docker -r`  
   Or option `-p|--plot` to plot graphs  
   e.g. `docker run -v /folder/for/results:/adb/tmp/ adb-docker -p --save`

tomjaguarpaw:
This looks very nice so far @iliaeg!

msdmkats:
I tried to use this and ran into two issues:
1. On windows git by default checks out files with CRLF line endings. When those get copied into a docker container, shell script simply crashes with strange diagnostic "file not found". I think, we should clone from the local repo instead of copying. Maybe make the cloned branch configurable.
2. After I forced LF line endings, run-all started and successfully computed manual for GMM and BA, but all hand problems crashed. Needs further investigation.

For the experiments I used a clean VM with Win Server 2019.

iliaeg:
> 1. On windows git by default checks out files with CRLF line endings. When those get copied into a docker container, shell script simply crashes with strange diagnostic "file not found". I think, we should clone from the local repo instead of copying. Maybe make the cloned branch configurable.

docker build performs COPY stage and all the commands after it every time again, so in my opinion it's more logical to do copy instead of git clone, so you can redo only commands that you need. To solve problem with line endings i can use `sed -i 's/\r//' run-wrapper.sh`.
On the other hand I can make two dockerfiles one to build base image with all required linux packages and use it as base image for second docker where would be git clone and cmake stage which we must build with `--no-cache flag`. 

> 2. After I forced LF line endings, run-all started and successfully computed manual for GMM and BA, but all hand problems crashed. Needs further investigation.

Im not sure when this problem appeared, i ran master branch under wsl and got just the same result for manual hand:
`WARNING: Command ran, but did not produce output file...`
I'll try to find out what is happening.

msdmkats:
> > 1. On windows git by default checks out files with CRLF line endings. When those get copied into a docker container, shell script simply crashes with strange diagnostic "file not found". I think, we should clone from the local repo instead of copying. Maybe make the cloned branch configurable.
> 
> docker build performs COPY stage and all the commands after it every time again, so in my opinion it's more logical to do copy instead of git clone, so you can redo only commands that you need. To solve problem with line endings i can use `sed -i 's/\r//' run-wrapper.sh`.

There're all the other files that may or may not cause a similar problem.
I guess, we can start with sed and see if there will be any other failures.

> > 1. After I forced LF line endings, run-all started and successfully computed manual for GMM and BA, but all hand problems crashed. Needs further investigation.
> 
> Im not sure when this problem appeared, i ran master branch under wsl and got just the same result for manual hand:
> `WARNING: Command ran, but did not produce output file...`
> I'll try to find out what is happening.

Hm-m... On the line 8 of `src/cpp/runner/HandBenchmark.cpp` I see a `\` being used as a path separator. Unfortunately, I can't check if that is _the_ problem right now.

msdmkats:
> Hm-m... On the line 8 of `src/cpp/runner/HandBenchmark.cpp` I see a `\` being used as a path separator. Unfortunately, I can't check if that is _the_ problem right now.

Yep. Just checked - this is the problem. After I replaced `\\` with `/`, runner worked.

tomjaguarpaw:
That's reassuring!

iliaeg:
> > Hm-m... On the line 8 of `src/cpp/runner/HandBenchmark.cpp` I see a `\` being used as a path separator. Unfortunately, I can't check if that is _the_ problem right now.
> 
> Yep. Just checked - this is the problem. After I replaced `\\` with `/`, runner worked.

Fine! I'll add commit with fix

msdmkats:
And, @iliaeg can you, please, add an option to run `ctest` in the container, preferably, in a way that would propagate the `ctest`'s exit code to the caller of docker.

iliaeg:
> And, @iliaeg can you, please, add an option to run `ctest` in the container, preferably, in a way that would propagate the `ctest`'s exit code to the caller of docker.

done

msdmkats:
I can confirm, that now on Windows Server 2019 VM, docker container builds, tests run and pass (excluding PyTorch hand - fix is in another branch). I'm in a process of doing a full run now.

tomjaguarpaw:
This looks very nice.  Is it ready to merge?  As long as it runs correctly I am happy for it to be merged.

iliaeg:
> This looks very nice. Is it ready to merge? As long as it runs correctly I am happy for it to be merged.

Docker for cpu version of python tools runs correctly, but we have a problem that run-wrapper does not return an error code in case of run-all.ps1 fails.
Also there is a problem with tensorflow-gpu, but I can make another PR when I figure out whats wrong.

iliaeg:
Docker cpu implemented. GPU support of python tools will be done later.

Issue #112, opened 2019-10-02
PyTorch Hand differentiation is wrong


mikhailnikolaev:
_PyTorch_ calculates jacobian of **Hand** objective incorrectly. In its result derivatives by several last _thetas_ variables are always zero, but in fact they aren't. By the way, **Hand** objective itself is calculated correctly.

mikhailnikolaev:
Closed due to PR #135

PR #135, opened 2019-10-19
Repairing PyTorch Hand objective differentiation

mikhailnikolaev:
Fixings that are applied due to issue #112 

Additional changes:
1. Hand test data flattening for Python
2. Decreasing tolerance for PyTorch to 1e-7 because precision 1e-8 is too small for Hand objective differentiation in complicated case for some sets of input data
3. Some code fixings and refactoring

# Graphs
HAND (Simple, Small)  Jacobian  - Release Graph


mikhailnikolaev:
> That looks great! The code became simpler 👍

Thanks!

Issue #126, opened 2019-10-11
BA and Hand plot wrong X axis


mikhailnikolaev:
I've realized what is the problem with graphs for BA and Hand objectives. First of all, X axis has to contain information about input variable count. So, now in the script `ADBench/plot_graphs.py` information about run variable count is extracted from the time file name (respected functions are defined in  `ADBench/utils.py`). File names for GMM and LSTM objectives contain needed information about parameter count (e.g. the typical LSTM time file has the next form `lstm_l2_c1024_times.txt`, so, the script can understand that there was 2 * 1024 variables). But BA and Hand file names have the folowing form:
```
<tool_name><number in order>.txt
```
and for such files the script uses the standard extracting logic, actually, it just concatenates all digits from the file name. Note, that this file name pattern is just taken from data files, stored in the directory `data`.

So, I think that the script `ADBench/run_all.ps1` must generate files for BA and Hand using the same pattern as it is provided for LSTM and GMM. That means that the script has to genrate new file names for result that in general case can be different from data input file names.

Another option is just leaving things as they are. Eventually, we just want to compare tools by their efficiency, I think, we don't need to know how many variables were used.

 What do you think about this?

awf:
Can't we just rename the BA data files?

mikhailnikolaev:
Hmmm... It's strange that I haven't thought about such an obvious solution. Yes, I think it would be the best decision in this situation

awf:
No worries :) Sometimes one is too close to the problem :)

mikhailnikolaev:
Closed due to PR #137

PR #105, opened 2019-09-27
Python runner and PyTorch module implementation

mikhailnikolaev:
1) Python runner is implemented.
2) PyTorch module for python is implemented. 
     **Known issue:** The new version of PyTorch (as well as the old one) produces wrong jacobian for _Hand_. Respective issue with detail description will be opened after PR completion.
3) Unit tests for python runner and common unit tests for all python modules are implemented.
4) Old realization of PyTorch benchmark is removed.

# Graphs for PyTorch
**Note:** for these graphs maximum timeout is 300 seconds.
**Note:** the old version of PyTorch is added to the graphs _**GMM**_ and _**LSTM**_ (its name is  _PyTorchOld_). For _**BA**_ and _**Hand**_ the old version crashed/terminated for all dots, so, it is not added to such graphs.

## Jacobian
BA  Jacobian  - Release Graph
GMM (1k)  Jacobian  - Release Graph
GMM (10k)  Jacobian  - Release Graph
HAND (Complicated, Big)  Jacobian  - Release Graph
HAND (Complicated, Small)  Jacobian  - Release Graph
HAND (Simple, Big)  Jacobian  - Release Graph
HAND (Simple, Small)  Jacobian  - Release Graph
LSTM  Jacobian  - Release Graph

##  Jacobian ÷ objective
HAND (Simple, Big)  Jacobian ÷ objective  - Release Graph
HAND (Simple, Small)  Jacobian ÷ objective  - Release Graph
LSTM  Jacobian ÷ objective  - Release Graph
BA  Jacobian ÷ objective  - Release Graph
GMM (1k)  Jacobian ÷ objective  - Release Graph
GMM (10k)  Jacobian ÷ objective  - Release Graph
HAND (Complicated, Big)  Jacobian ÷ objective  - Release Graph
HAND (Complicated, Small)  Jacobian ÷ objective  - Release Graph

## Objective
BA  Objective  - Release Graph
GMM (1k)  Objective  - Release Graph
GMM (10k)  Objective  - Release Graph
HAND (Complicated, Big)  Objective  - Release Graph
HAND (Complicated, Small)  Objective  - Release Graph
HAND (Simple, Big)  Objective  - Release Graph
HAND (Simple, Small)  Objective  - Release Graph
LSTM  Objective  - Release Graph

##  Objective ÷ manual
HAND (Complicated, Big)  Objective ÷ manual  - Release Graph
HAND (Complicated, Small)  Objective ÷ manual  - Release Graph
[...]

PR #137, opened 2019-10-20
BA and Hand graph X-axis correcting

mikhailnikolaev:
Related to issue #126.

X-axes for BA and Hand graphs are corrected. Now for BA X-axis shows all scalar variable count (11n + 3m + 2p) and for Hand X-axis shows count of correspondencies. This info is extracted from data test file names, so, such files are renamed with adding respected postfixes.

# Graph examples
## Static
BA  Jacobian  - Release Graph
HAND (Simple, Small)  Jacobian  - Release Graph

## Plotly
image
image


msdmkats:
If x-axis for hand shows the number of correspondences, then, maybe we should make it say so? Because the number of independent variables is `26` for simple cases and `26 + 2 * n_correspondences` for complicated ones.

mikhailnikolaev:
> If x-axis for hand shows the number of correspondences, then, maybe we should make it say so? Because the number of independent variables is `26` for simple cases and `26 + 2 * n_correspondences` for complicated ones.

It looks reasonable. I've done this

tomjaguarpaw:
I'm not familiar with HAND, but does Manual really not get slower for bigger problem sizes?

mikhailnikolaev:
> I'm not familiar with HAND, but does Manual really not get slower for bigger problem sizes?

It does, but in relation to others tools slowdown doesn't look very significant. I mean, it's very hard to notice deceleration in such a scale. You can look at plots from PR #105, for example. You will see the same situation, but that PR was closed about a month ago, so, there is not possibility that the current PR could break graph plotting for Manual HAND.

PR #139, opened 2019-10-22
Tensorflow Hand module

mikhailnikolaev:
**This PR is draft because graphs for it are being calculated**
Tensorflow Hand objective differentiation test class is implemented.

__Note__: Tensorflow 2.0 is used. All calculations are made in eager execution mode.


mikhailnikolaev:
@awf, @tomjaguarpaw, I've moved us jacobian part merging into two columns to `output` function of the test class. New plots are attached to the PR. I attach old plots to this message for comparision. Note, that I've merged PyTorch new Hand version to the branch, so, little changes in PyTorch plot behaviour are correct. Also, new X-axis for Hand is used. As you can see there are not any significant changes in time for Tensorflow after refactoring. But I like the new version of the code, because it becames clearer, so, I've added the commit with changes to the PR.

# Plots
**Note**: there are only graphs for complicated jacobian calculating, because only these plots could be changed.
## Jacobian

## Jacobian ÷ objective
HAND (Complicated, Big)  Jacobian ÷ objective  - Release Graph
HAND (Complicated, Small)  Jacobian ÷ objective  - Release Graph

tomjaguarpaw:
Good, well I'm happy with it.  I think we should merge it.  If @awf has comments then we can address them post-merge.

PR #140, opened 2019-10-23
Add line ending after writing times and value

tomjaguarpaw:
I don't know any reason why it would break things to add a new line, but having them improves the user experience of trying to view the files.


tomjaguarpaw:
I tested it and it seems to work, so merging ...

PR #141, opened 2019-10-24
Violation info fixing

mikhailnikolaev:
Now graph plotting script counts only violations that are not related to timeouts. Later in cases of timeout handle for violation markers was added even if there wasn't any calculation violations.


tomjaguarpaw:
Thanks for this.  I'll have a look on Monday.

tomjaguarpaw:
Looks good but can we improve the naming?  Perhaps `violation_x` can become `non_timeout_violation_x` instead.  Since it is only a set of the violations it would be better to make that clearer in the name.

tomjaguarpaw:
(Sorry for my late response to this too)

mikhailnikolaev:
> Looks good but can we improve the naming? Perhaps `violation_x` can become `non_timeout_violation_x` instead. Since it is only a set of the violations it would be better to make that clearer in the name.

Done

PR #144, opened 2019-10-28
ADBench web view application

mikhailnikolaev:
Web App for run result plot (static and plotly) viewing is implemented.
This application has been published on Azure. Its URL is https://adbenchwebviewer.azurewebsites.net

**Note**: in some web browsers (e.g. Google Chrome) plotly HTML pages are downloaded instead of displaying. That is because HTML pages stored in the blob container by Azure Batch have content type *octet-stream* and the property *type* of the tag **\<a\>** in web app response could be ignored be the browser.

### DIFF
--- a/ADBench/ADBenchWebViewer/ADBenchWebViewer/Controllers/PlotsController.cs
+++ b/ADBench/ADBenchWebViewer/ADBenchWebViewer/Controllers/PlotsController.cs
@@ -4,24 +4,24 @@ namespace ADBenchWebViewer.Controllers
 {
     public class PlotsController : Controller
     {
-        private readonly IRunInfoProvider shareHelper;
+        private readonly IRunInfoProvider runInfoProvider;
 
-        public PlotsController(IRunInfoProvider shareHandler) => this.shareHelper = shareHandler;
+        public PlotsController(IRunInfoProvider runInfoProvider) => this.runInfoProvider = runInfoProvider;
 
         public IActionResult Index()
         {
-            var dirInfo = shareHelper.GetRunsInfo();
+            var dirInfo = runInfoProvider.GetRunsInfo();
             return View(dirInfo);
         }
 
         public IActionResult GetStatic(string dir)
         {
-            return View(shareHelper.GetPlotsInfo(dir));
+            return View(runInfoProvider.GetPlotsInfo(dir));
         }
 
         public IActionResult GetPlotly(string dir)
         {
-            return View(shareHelper.GetPlotsInfo(dir));
+            return View(runInfoProvider.GetPlotsInfo(dir));
         }
     }
 }
--- a/ADBench/ADBenchWebViewer/ADBenchWebViewer/Controllers/PlotsController.cs
+++ b/ADBench/ADBenchWebViewer/ADBenchWebViewer/Controllers/PlotsController.cs
@@ -1,0 +1,27 @@
+﻿using Microsoft.AspNetCore.Mvc;
+
+namespace ADBenchWebViewer.Controllers
+{
+    public class PlotsController : Controller
+    {
+        private readonly IRunInfoProvider shareHelper;
+
+        public PlotsController(IRunInfoProvider shareHandler) => this.shareHelper = shareHandler;
+
+        public IActionResult Index()
+        {
+            var dirInfo = shareHelper.GetRunsInfo();
+            return View(dirInfo);
+        }
+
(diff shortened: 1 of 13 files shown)


ADBench/ADBenchWebViewer/ADBenchWebViewer/Controllers/PlotsController.cs
```diff
@@ -0,0 +1,27 @@
+﻿using Microsoft.AspNetCore.Mvc;
+
+namespace ADBenchWebViewer.Controllers
+{
+    public class PlotsController : Controller
+    {
+        private readonly IRunInfoProvider shareHelper;
```
msdmkats: Consider renaming `shareHelper -> runInfoProvider`

mikhailnikolaev: Done

tomjaguarpaw:
Is this a viewer to view static files?

tomjaguarpaw:
The view of all the graphs looks very nice.  It seems to me, though, that the application is generating the pages that hold the graphs.  Is that right?  If so, couldn't they be static too?

mikhailnikolaev:
> Is this a viewer to view static files?

No, this is a viewer both for plotly and for png files, generated by plot script

mikhailnikolaev:
> The view of all the graphs looks very nice. It seems to me, though, that the application is generating the pages that hold the graphs. Is that right? If so, couldn't they be static too?

Yes, this application generates the pages with graphs (both plotly and static) that are stored in the specific Azure Storage. Further this storage will be populated by weekly run process that will calculate values for all tools and will create respected graphs. So, such pages can't be static due to information population from week to week

tomjaguarpaw:
Ah, maybe my question was ambiguous.  Which pages on https://adbenchwebviewer.azurewebsites.net/ are generated by a server process on each access and which are immutable pages in the blob storage bucket?

mikhailnikolaev:
> Ah, maybe my question was ambiguous. Which pages on https://adbenchwebviewer.azurewebsites.net/ are generated by a server process on each access and which are immutable pages in the blob storage bucket?

All pages are generated dynamically by the server

tomjaguarpaw:
Ah, I see.  Could we generate the pages offline instead then?  I think maintaining a server is liable to be costly in terms of time.

msdmkats:
First, let me clarify what's static and what's dynamic here:
Static: png images and htmls generated by plotly. They reside in Azure storage.
Dynamic:
- A page that list experiment runs (a run produces a full set of graphs) - effectively lists folders in a certain container in Azure storage
- A page that lists and links to plotly graphs for a given run in Azure storage
- A page that lists and renders via hotlinking png graphs for a given run in Azure storage

A few notes about server:
- It actually is optional: there're no credentials involved, because the container in Azure storage allows public read access. So you can just clone the repo and run the web-app on your machine. It will function just like one we've deployed and show the contents of the same Azure storage container.
- We host it in an Azure App Service, so there's no server maintenance involved. Redeployment in case of update can be done from Visual Studio.

And finally, here's why I'm strongly for the dynamic app and against static pages:
- I can't really imagine anything that could go wrong with the app, that couldn't with static pages - said pages would probably have to be hosted somewhere too.
- If the static pages are produces by the same process that weekly builds all graphs, then the very same process will have to upload the generated page to where it should be hosted, which is a point of potential failure, and that's after two days of computations.
- It's really easy for static pages to go out of sync with the storage:
  - Any manual changes to the contents of the storage (add or delete a run or, especially, a part of a run) will either be very painful or cause a desynchronization.
  - Moving to another storage account would invalidate all static pages immediately.
  - Any design changes would be a great pain to apply to existing pages.

tomjaguarpaw:
Even a very simple Azure App Service application will require someone's time and effort to attend to if it goes wrong.  Something as small as having to go to the Azure Portal and click restart can be a major annoyance for someone who has many other pressing things to attend to.  From there it only gets more complicated.  Redeployment from Visual Studio is only convenient for someone who is already familiar with Visual Studio (and I am not).

My ideal is that all pages of this site (except perhaps the very front page) are static and immutable.  Once they are deployed they never change.  If we ever want to change the design of the pages then we generate entirely new pages, upload them with entirely new names, and change only the link on the front page (this is not too different from how git operates).

This system would also allow site generation to be entirely decoupled from graph generation.

Issue #143, opened 2019-10-24
Why do we add 2 in LSTM logsumexp?


tomjaguarpaw:
LSTM logsumexp has [2 added to the sum](https://github.com/awf/ADBench/blob/5020acbcd3948aab74be042a9cbaf670bb28b21c/src/cpp/shared/lstm.h#L22).  Do we know why this is?  The earliest occurrence I can see is https://github.com/awf/ADBench/commit/d75f713d8a6fa4f8ff3f988692c1a5c76238abe9#diff-cadabfe323e7f3ac5fe22b5de628d2aeR24

awf:
Yes that is very odd, and doesn't ring a bell.   Let's also make this issue track other LSTM oddnesses: diagonal weight matrix.  Any more?

PR #147, opened 2019-10-30
Special-case display of "LSTM" name

tomjaguarpaw:
because of https://github.com/awf/ADBench/issues/143

Eventually we plan to change to a standard LSTM formulation

### DIFF
--- a/ADBench/plot_graphs.py
+++ b/ADBench/plot_graphs.py
@@ -76,7 +76,7 @@ def graph_data(build_type, objective, maybe_test_size, function_type):
 
     test_size = ", ".join([utils.cap_str(s) for s in maybe_test_size[0].split("_")]) if len(maybe_test_size) == 1 else None
     has_ts = test_size is not None
-    graph_name = (f"{objective.upper()}" +
+    graph_name = (f"{objective_display_name(objective)}" +
                   (f" ({test_size})" if has_ts else "") +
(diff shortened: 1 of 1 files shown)


ADBench/plot_graphs.py
```diff
@@ -85,6 +85,18 @@ def graph_data(build_type, objective, maybe_test_size, function_type):
...
 
+# What we call LSTM is not quite a full LSTM.  Rather it's an LSTM
+# with diagonal weight matrices.  We don't want to be misleading so we
+# rename the graph.  Eventually we will implement a full LSTM and we
+# will remove this special case.  See
+#
+#     https://github.com/awf/ADBench/issues/143
+def objective_display_name(objective):
+    if objective.upper() == "LSTM":
+        return "D-LSTM"
```
mikhailnikolaev: What does the "D" mean?

tomjaguarpaw: "Diagonal"

awf:
I'm keen on D-LSTM, we could also use IndyLSTM as per recent Google paper, but this was coded before that.

Sent on a mobile device, please excuse brevity and typos. I'll generally try to follow up when I get to a keyboard.

Andrew Fitzgibbon
Microsoft
________________________________

tomjaguarpaw:
Yes, let's just go with D-LSTM for now.  It's easy enough to change it again later if we need.

PR #148, opened 2019-10-31
PR Build Definitions

msdmkats:
Definitions for PR builds and fixes to issues that were found during their creation.

Builds:
- `adobuilds/pr-win.yml` - build with cmake/visual studio, run ctest, run a subset of benchmarks, output results as artifacts
- `adobuilds/pr-linux-docker.yml` - build docker container, run ctest, run a subset of benchmarks, output results as artifacts.

Examples of successful and failed (when they should) builds can be found
- `adobuilds/pr-win.yml` - https://msrcambridge.visualstudio.com/Knossos/_build?definitionId=341
- `adobuilds/pr-linux-docker.yml` - https://msrcambridge.visualstudio.com/Knossos/_build?definitionId=340

New functions introduced:
- `run-all.ps1` can delete Jacobians when they are found to be correct to save space
- `run-all.ps1` produces a non-zero exit code when some of the benchmarks crash, timeout, or produce incorrect results
- `run-all.ps1` allows all parameters affecting which benchmarks are run to be configured by the caller
- dotnet tests produce trace files

Issues that were fixed:
- dotnet tests were run in parallel despite `ModuleLoader` not being thread-safe. That caused transient failures.
- Tests for Julia BA, Hand, and LSTM benchmarks were not run by ctest
- Tests for dotnet benchmarks were not run by ctest
- `Autograd_ba.py` file was named `autograd_ba.py` (notice case) which cased failures in run-all on linux


ADBench/run-all.ps1
```diff
@@ -53,23 +53,51 @@ param(# Which build to test.
...
+      [string[]]$gmm_sizes = @("1k", "10k"),
+
+      # Hand problem sizes to try. Must be a subset of @("small", "big")
+      [string[]]$hand_sizes = @("small", "big"),
+
+      # Number of the first BA problem to try. Must be between 1 and ba_max_n
+      [int]$ba_min_n = 1,
+
+      # Number of the last BA problem to try. Must be between ba_min_n and 20
+      [int]$ba_max_n = 5,
```
tomjaguarpaw: It would seem more natural to me if this was the largest possible (which seems to be 20).  Is there a good reason the default wouldn't be that?

ADBench/run-all.ps1
```diff
@@ -53,23 +53,51 @@ param(# Which build to test.
...
+
+      # Number of the first Hand problem to try. Must be between 1 and hand_max_n
+      [int]$hand_min_n = 1,
+
+      # Number of the last Hand problem to try. Must be between hand_min_n and 12
+      [int]$hand_max_n = 5,
```
tomjaguarpaw: It would seem more natural to me if this was the largest possible (which seems to be 12).  Is there a good reason the default wouldn't be that?

ADBench/run-all.ps1
tomjaguarpaw: (I see that the original default was 5.  If the max possible is 20 wouldn't that be a better default?  Maybe we should leave that for another PR though.)

ADBench/run-all.ps1
tomjaguarpaw: [See other comment for related discussion]

ADBench/run-all.ps1
```diff
@@ -119,6 +172,7 @@ function run_command ($indent, $outfile, $timeout, $cmd) {
     } else {
         $Process.Kill()
         Write-Host "${indent}Killed after $timeout seconds"
+        Store-NonFatalError "Process killed after $timeout seconds`n[$cmd $args]"
```
tomjaguarpaw: Might it be better to use `Report-NonFatalError` here and get rid of the `Write-Host` above?

ADBench/run-all.ps1
```diff
@@ -363,7 +408,13 @@ Class Tool {
             $comparison.CompareFiles($current_jacobian_path, $golden_jacobian_path)
```
tomjaguarpaw: This is a question not directly about this PR but reading these changes just made me think of it: do we not compare the objective values too?  It looks like we're only comparing the Jacobian.

ADBench/run-all.ps1
tomjaguarpaw: (I never knew that BA and HAND could be bigger and I'm quite surprised to discover it!)

tomjaguarpaw:
@msdmkats: Several months ago I wrote [azure-devops.yml](https://github.com/awf/ADBench/blob/a471132814acfa53a8b5d158931b90353fb4785d/azure-devops.yml) and [azure-devops-ci.ps1](https://github.com/awf/ADBench/blob/a471132814acfa53a8b5d158931b90353fb4785d/ADBench/azure-devops-ci.ps1) to do a more rudimentary version of this PR.  I would appreciate it if you would take a look at them and determine whether your PR does everything that they do, so that they could be deleted.  I hope they can!

ADBench/run-all.ps1
msdmkats: While the best person to explain the justification behind the current defaults is, probably, @awf , here're some of my thoughts on the matter:

20 would be a bad default for BA, because that's the one where just the size of the output Jacobian is around 20 GB. That would simply crash most tools on most machines.
Even before that the sizes of the result quickly become very big to the point, where just the time to save them could cause the timeout. So, it's important to balance this default and the default for timeout.

We also need to balance these defaults against the time needed to complete the default set of benchmarks. It's two days as it is.

I think, it would be best to raise an issue about the defaults, and change them in a separately, if we decide to do so.

Anyway, the defaults are not the topic of this PR.

ADBench/run-all.ps1
msdmkats: See above.

ADBench/run-all.ps1
msdmkats: I don't think so. Non-fatal errors are printed again at the end of the script, so that the user wouldn't have to read through the whole log to check if they occurred. `"${indent}Killed after $timeout seconds"` won't be a useful error message there, because it lacks necessary context (what was killed) by itself. So I put this context into the stored message. However when we kill the process, all the context was just printed into the terminal and it won't make much sense to repeat it. Hence the different texts to print when the error just occurred and in the end.

ADBench/run-all.ps1
msdmkats: Yes, we compare only the Jacobians. Comparing the objectives themselves would bring some more complexity to run-all, plot-graphs, and JacobianComparisonLib (BA objective is non-trivial), without bringing much value - correct Jacobians mean that the implementation of the objective is off by constant at most, and it generally wasn't produced by the AD tool, so that's not a problem. If the Jacobian is off, checking, whether the the objective is correct would be the first step anyway.

msdmkats:
> @msdmkats: Several months ago I wrote [azure-devops.yml](https://github.com/awf/ADBench/blob/a471132814acfa53a8b5d158931b90353fb4785d/azure-devops.yml) and [azure-devops-ci.ps1](https://github.com/awf/ADBench/blob/a471132814acfa53a8b5d158931b90353fb4785d/ADBench/azure-devops-ci.ps1) to do a more rudimentary version of this PR. I would appreciate it if you would take a look at them and determine whether your PR does everything that they do, so that they could be deleted. I hope they can!

My PR runs a smaller subset of benchmarks, if that's not a problem, we can delete the older files. Is it?

ADBench/run-all.ps1
tomjaguarpaw: OK, I don't really mind.  I prefer less code to more but as you wish.

ADBench/run-all.ps1
tomjaguarpaw: OK

ADBench/run-all.ps1
tomjaguarpaw: > Anyway, the defaults are not the topic of this PR.

Agreed

ADBench/run-all.ps1
tomjaguarpaw: > correct Jacobians mean that the implementation of the objective is off by constant at most

Good point

tomjaguarpaw:
> @tomjaguarpaw My PR runs a smaller subset of benchmarks, if that's not a problem, we can delete the older files. Is it?

Do you mean only that it runs the tools on fewer x-axis points?  If so then that is not a problem.

Please don't delete them though.  I just want to know your opinion so I feel more confident deleting them in the future.

msdmkats:
> Do you mean only that it runs the tools on fewer x-axis points? If so then that is not a problem.

Exactly.

> Please don't delete them though. I just want to know your opinion so I feel more confident deleting them in the future.

OK.

tomjaguarpaw:
> >    Do you mean only that it runs the tools on fewer x-axis points? If so then that is not a problem.

> Exactly.

Excellent.  I sometimes use my CI build for running larger runs of quicker tools so I will keep it around until I learn a better way.

PR #149, opened 2019-11-01
Run all tool process in Azure Batch

mikhailnikolaev:
Azure Batch Job Schedule for weekly running all tools and creating graphs with storing them to the blob container, that can be observed by the [ADBench Web Viewer](https://github.com/awf/ADBench/pull/144), is created in [this](https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/0f64a630-b912-4501-ad57-262a9e12637c/resourceGroups/ADBench/providers/Microsoft.Batch/batchAccounts/adbenchrunbatch/accountOverview) Azure Batch Account. This PR adds info about this job schedule.

Adding files:

1. Bash script that is run in a pool node.
2. JSON file that describes job schedule with missed blob container URL. It can be used for creating a new Azure Batch Job Schedule
3. Documentation that describes the script and the weekly run process and also contains instructions about creating a new job schedule.

**Note**: this PR is partially connected with #144 (by the documentation), so, it should be merged after that PR will have completed.

**Possible improving**: maybe we need to increase timeout for this kind of tool run, e.g. up to 10 or even 15 minutes

### DIFF
--- a/ADBench/AzureBatch/runallscript.sh
+++ b/ADBench/AzureBatch/runallscript.sh
@@ -18,6 +18,8 @@ DOCKER_INSTALLATION_PROBLEM_EXIT_CODE=3
 DOCKER_ACTIVATING_PROBLEM_EXIT_CODE=4
 DOCKER_BUILD_PROBLEM_EXIT_CODE=5
 TEST_FAIL_EXIT_CODE=6
+TOOL_RUNNING_FAILURE_EXIT_CODE=7
+PLOT_CREATING_FAILURE_EXIT_CODE=8
 
 
 
@@ -118,8 +120,23 @@ mkdir tmp
 # Run all tools.
 docker run -v $(pwd)/tmp:/adb/tmp/ adb-docker -r
 
+# Check tool running.
+# Note: run-all.ps1 script returns "8" in case of non-fatal errors.
+if [[ $? -ne 0 && $? -ne 8 ]]
+then
+    echo "Running all tools failure! Stopping the script"
+    exit $TOOL_RUNNING_FAILURE_EXIT_CODE
+fi
+
 # Create plots.
 docker run -v $(pwd)/tmp:/adb/tmp/ adb-docker -p --save --plotly
+
+# Check plot creating.
+if [[ $? -ne 0 ]]
+then
+    echo "Plot creating failure! Stopping the script"
+    exit $PLOT_CREATING_FAILURE_EXIT_CODE
+fi
 
 # Create output directory name in the format:
 #   <year>-<month>-<day>_<hour>-<minute>-<second>_<commit hash>
--- a/ADBench/AzureBatch/runallscript.sh
+++ b/ADBench/AzureBatch/runallscript.sh
@@ -1,0 +1,113 @@
+#!/bin/bash
+
+# This script is created for Ubuntu 18.04 for Azure Batch.
+# It should be run by the root user.
+
+# Script constants.
+REPOSITORY_URL="https://github.com/awf/ADBench"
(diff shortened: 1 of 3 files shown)


ADBench/AzureBatch/runallscript.sh
```diff
@@ -0,0 +1,113 @@
...
+    echo "Docker is not active! Stopping the script"
+    exit $DOCKER_PROBLEM_EXIT_CODE
+fi
+
+# Prepare docker output directory.
+mkdir tmp
+
+# Create docker image, run all tools and create plots.
+docker build -t adb-docker . \
+    && docker run -v $(pwd)/tmp:/adb/tmp/ adb-docker -r \
```
msdmkats: I'd like to suggest invoke ctest before this: some tools/or interpreters (e.g. dotnet, julia) perform some long operations the first time they run, which theoretically can cause some unjustified timeouts. Running ctest before actual benchmarks would eliminate that ptoblem.

ADBench/AzureBatch/runallscript.sh
```diff
@@ -0,0 +1,113 @@
...
+    && docker run -v $(pwd)/tmp:/adb/tmp/ adb-docker -p --save --plotly
```
msdmkats: `run-all.ps1` now returns a non-zero exit code (specifically, 8) when a non-fatal error occurs, such as one of the tools crashing on one of the benchmarks. That shouldn't prevent graphs from being plotted and published.

ADBench/AzureBatch/runallscript.sh
mikhailnikolaev: Hmmm, it is strange but ok, I will add CTest running to the script

msdmkats: Nothing strange here. It's just that on our development machine we install our tools once and use them for a long time, forgetting the "first run experience", while on azure batch every time is the first time.

msdmkats:
About possible improvements to the process: we can not only increase the timeout, but also run more benchmarks for BA (default is 5 out of 20) and hand (default (5 out of 12) x 4) problems.

ADBench/AzureBatch/runallscript.sh
mikhailnikolaev: Changed

ADBench/AzureBatch/runallscript.sh
mikhailnikolaev: Fixed

docs/AzureBatch.md
```diff
@@ -2,43 +2,49 @@
...
 
 ### The script
-`runallscript.sh` is a bash script that is designed for *Ubuntu 18.04*. It has descriptive comments, so, you can find more information about its behaviour in them. The script returns the following exit codes:
+`/ADBench/AzureBatch/runallscript.sh` is a bash script that is designed for *Ubuntu 18.04*. It has descriptive comments, so you can find more information about its behaviour in them. The script returns the following exit codes:
```
tomjaguarpaw: Could we specify all paths from the top-level of the repository (and in particular not assume that the repository directory is called `ADBench`)?

mikhailnikolaev: As you can see, paths have already been specified from the repository top-level. `/ADBench/AzureBatch/runallscript.sh` is located in the directory `ADBench/AzureBatch` inside the repository root, so `ADBench` is not the root of the repository

tomjaguarpaw: Ah great, in that case I think it would be clearer to drop the leading `/` from the paths.

mikhailnikolaev: Done

PR #150, opened 2019-11-05
Plot creating documentation

mikhailnikolaev:
Documentation about `plot_graphs.py` is added


docs/PlotCreating.md
```diff
@@ -0,0 +1,115 @@
...
+## Plot Script Definition
+
+Special plotting script is responsible for creating benchmark plots. It is a Python3 script defined in the file `/ADBench/plot_graphs.py`. It has the following dependencies:
+
+* matplotlib
+* plotly
+
+These dependencies are fetched automatically during CMake configure.
+
+This script looks through the directory `/tmp` and for every subdirectory, that contains results produced by [the global runner](./GlobalRunner.md), it creates graphs and stores them in the folder `/tmp/graphs`. Two kinds of graphs are available: static PNG files and dynamic [plotly](https://plot.ly/python/) HTML files.
```
tomjaguarpaw: Isn't it better to say `tmp` rather than `/tmp` here and several other places?  `/tmp` generally implies (to a UNIX user) that `tmp` is in the root directory.

mikhailnikolaev: All paths in this file are changed to be relative to the project root

tomjaguarpaw:
Thanks for making that fix.  Looks very good to me now.

PR #158, opened 2019-11-13
Timeout for Azure Batch is doubled

mikhailnikolaev:
**Note**: the new version of the script has already been added to the [Azure Batch](https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/0f64a630-b912-4501-ad57-262a9e12637c/resourceGroups/ADBench/providers/Microsoft.Batch/batchAccounts/adbenchrunbatch/accountOverview)


tomjaguarpaw:
Yes, as I understand it that matches what @awf wants, but best to wait for him.

PR #159, opened 2019-11-24
Guaranteed timeout checking

mikhailnikolaev:
Global runner script checks guaranteed timeouts now. That means that tests of bigger sizes won't be run if tests of smaller sizes are interrupted due to timeout.


ADBench/run-all.ps1
```diff
@@ -329,21 +350,21 @@ Class Tool {
     }
 
     # Run a single test
-    [void] run ([string]$objective, [string]$dir_in, [string]$dir_out, [string]$fn) {
-        if ($objective.contains("Eigen")) { $out_name = "$($this.name)_Eigen" }
-        elseif ($objective.endswith("SPLIT")) { $out_name = "$($this.name)_split" }
-        else { $out_name = $this.name }
-        $output_file = "${dir_out}${fn}_times_${out_name}.txt"
+    # Returns true if there was a timeout and false otherwise
+    [bool] run ([string]$objective, [string]$dir_in, [string]$dir_out, [string]$fn) {
```
msdmkats: Usually, returned `true` value indicates success, and `false` some sort of error. That being inversed may be confusing, especially since the function is called `run` and not something like `did_timeout`.

msdmkats: Maybe, define a RunResult enumeration with values like `Success`, `Timeout`, `Skipped`, `ResultIncorrect`, etc. and return it?

mikhailnikolaev: Yes, I agree with this. Changes are performed

docs/GlobalRunner.md
```diff
@@ -20,6 +20,8 @@ For every specified _testing module_ `run-all.ps1` will run every specified benc
...
 
+The script checks certain timeouts. This means that benchmark for the bigger test will not be run if the run test for the smaller size finishes with timeout. Thus, if GMM objective benchmark test is terminated due to timeout in case of `K = 10` and some `d` then tests for bigger `K` values will not be run (both for the current `d` value or for bigger `d` values), but for `1k` and `10k` such a checking is independent. For LSTM certain timeout control is the same (but for `l` and `c` instead of `d` and `K` respectively). For BA and Hand test with bigger order number will not be run if any test with less number is terminated due to timeout. So, for them tests are expected to be sorted respective their sizes.
```
tomjaguarpaw: I suggest making this description more symmetrical between `d` and `K` (it's only an implementation detail that `K` is the inner loop and `d` the outer).  Perhaps something like "Thus, a GMM objective test will not be run in any case where the `d` and `K` values are both greater than or equal to the `d` and `K` values of a previously timed-out test."

tomjaguarpaw:
Ideally I would like to find a different word than "certain".  It's a bit ambiguous since it can also mean something more like "particular", as in "certain people have been working on the ADBench project".  I don't have a particularly compelling suggestion but "sure" and "guaranteed" are probably OK.

ADBench/run-all.ps1
```diff
@@ -147,7 +155,11 @@ function mkdir_p($path) {
    }
 }
 
+# Type of run command function returning value
+enum RunCommandStatus { Finished; Timeout }
+
 # Run command and (reliably) get output
+# Returns true if there was a timeout and false otherwise
```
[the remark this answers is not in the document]
mikhailnikolaev: Yes, it's an incorrect comment. Fixed

mikhailnikolaev:
> This looks like it will be very useful! Do we know how much this speeds up a full run?

Actually, we don't know yet, because full run is performed once a week. So, I hope this PR will finish this week till friday, thus the new full run will use a new version of the global runner.

docs/GlobalRunner.md
mikhailnikolaev: Changed

mikhailnikolaev:
> Ideally I would like to find a different word than "certain". It's a bit ambiguous since it can also mean something more like "particular", as in "certain people have been working on the ADBench project". I don't have a particularly compelling suggestion but "sure" and "guaranteed" are probably OK.

Certain timeout now is called guaranteed timeout

PR #160, opened 2019-11-30
Using Sort-Object

mikhailnikolaev:
This PR is done due to some bug in the current array parameter sorting.

A `sort` alias for the comandlet `Sort-Object` was used in the PowerShell code. It works fine in Windows, but doesn't in Linux, because there is an utility called `sort` in Linux. That's why in _PowerShell Core_ there is no alias `sort` for `Sort-Object`, so calling `sort` the script uses Linux utility `sort`, that sorting values alphabetically by default. It leads to some strange behaviour, for example the script calculates GMM for `d = 10` before `d = 2`. So, now the alias using is removed.

### DIFF
--- a/ADBench/run-all.ps1
+++ b/ADBench/run-all.ps1
@@ -103,10 +103,10 @@ param(# Which build to test.
 
 # Sort array parameters define test sizes ascending
 function sort_size_parameters() {
-    $script:gmm_d_vals = $script:gmm_d_vals | sort
-    $script:gmm_k_vals = $script:gmm_k_vals | sort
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
Interesting and well spotted.  How did this affect the weekly run?

mikhailnikolaev:
> Interesting and well spotted. How did this affect the weekly run?

You can see the result on http://adbenchwebviewer.azurewebsites.net/ In short, this leads to unpropper guaranteed timeout prediction. E.g. if GMM `K=100` is finished due to a timeout, then GMM `K=5, 25, 500` will not be run, because alphabetically they follow `100`

PR #163, opened 2019-12-06
GMM d128 data

iliaeg:
Data files for gmm with 128 dimensions are generated


awf:
Great to have these, thanks!

PR #165, opened 2019-12-14
Remove unused extensions

### LINKED ISSUE #164: VSCode extensions not found?
Missing extensions reported in VSCode, clean checkout, then "code ." in top level folder.

image

awf:
Fixes #164

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -6,8 +6,6 @@
     "recommendations": [
         "ms-vscode.cpptools",                   // C/C++ language support
         "twxs.cmake",                           // CMake language support
-        "vector-of-bool.cmake-tools",           // configure + build workflow for CMake-based projects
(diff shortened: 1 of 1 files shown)


.vscode/extensions.json
```diff
@@ -6,8 +6,6 @@
     "recommendations": [
         "ms-vscode.cpptools",                   // C/C++ language support
         "twxs.cmake",                           // CMake language support
-        "vector-of-bool.cmake-tools",           // configure + build workflow for CMake-based projects
```
msdmkats: Instead of removing completely, we should update the name - apparently the extension is now maintained by Microsoft and goes by the name `ms-vscode.cmake-tools`.

I also suggest just commenting out the second extension, until its maintainers produce a suitable update.

awf:
May be stale.   Others to repro #164 and fix.

PR #167, opened 2019-12-18
Say "timed out" instead of "terminated"

tomjaguarpaw:
because the former is more precise


tomjaguarpaw:
What do we think about saying "timed out" instead of "terminated"?  Whenever I see the latter I instinctively think "something else terminated it for an unknown reason" whereas I think, other than the crashes, this only happens in the case of timeout doesn't it?

PR #171, opened 2019-12-20
Accelerating PyTorch GMM objective

mikhailnikolaev:
GMM objective for PyTorch is refactored with accelerating.

**Note**: plots for the tools other than `Tensorflow`, `Manual` and `PyTorch` can be wrong.

# Examples of the old graphs
GMM (1k)  Jacobian  - Release Graph

# Examples of the new graphs
GMM (1k)  Jacobian  - Release Graph

### DIFF
--- a/src/python/modules/PyTorch/gmm_objective.py
+++ b/src/python/modules/PyTorch/gmm_objective.py
@@ -58,7 +58,7 @@ def constructL(d, icf):
 
 def Qtimesx(Qdiag, L, x):
 
-    f = torch.einsum('rbq,prq->prb', L, x)
+    f = torch.einsum('ijk,mik->mij', L, x)
     return Qdiag * x + f
 
 
--- a/src/python/modules/PyTorch/gmm_objective.py
+++ b/src/python/modules/PyTorch/gmm_objective.py
@@ -9,6 +9,14 @@ def logsumexp(x):
     return torch.log(sum(emx)) + mx
 
 
+def logsumexpvec(x):
+    '''The same as "logsumexp" but calculates result for each row separately.'''
+
+    mx = torch.max(x, 1).values
+    lset = torch.logsumexp(torch.t(x) - mx, 0)
+    return torch.t(lset + mx)
+
+
 def log_gamma_distrib(a, p):
     return scipy_special.multigammaln(a, p)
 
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
Wow, that is very impressive.

src/python/modules/PyTorch/gmm_objective.py
```diff
@@ -20,52 +28,54 @@ def sqsum(x):
...
     return torch.stack(columns, -1)
 
 
 def Qtimesx(Qdiag, L, x):
-    res = Qdiag * x
-    for i in range(L.shape[0]):
-        res = res + L[:, i] * x[i]
-    return res
 
+    f = torch.einsum('rbq,prq->prb', L, x)
```
awf: Is this idiomatic?  I would have thought ijk,mik->mij

mikhailnikolaev: Not, it isn't. I think your option is more convinient

awf:
Folks, I wonder if we might imagine including a version of the older code with a title like "PyTorch (naive)".   Just as we have a few C++ variants, we could imagine a couple of PyTorch variants.

tomjaguarpaw:
Yes, I think that sounds like a good idea.

mikhailnikolaev:
> Folks, I wonder if we might imagine including a version of the older code with a title like "PyTorch (naive)". Just as we have a few C++ variants, we could imagine a couple of PyTorch variants.

But it is still the PyTorch tool, but written in more effective way. Having "naive" and usual PyTorch tools can lead to misunderstanding what the "naive" version stands for, because it is just less effective code, not more than that. E.g. I can write much less effective code, will it be another version of PyTorch ("very naive") or not? In C++ libraries we use (as far as I understand) different libraries (e.g. Eigen) or different ways of using of them (e.g. Eigen and EigenVector), but here we use definetly the same tool and the same scenario. I don't think it should be show as another tool

awf:
I think the analogy to "EigenVector" and "Eigen" is appropriate.  In EigenVector, the author was trying to use the vectorization capabilities of Eigen, in a way that made sense in 2015, but today makes little difference.  Similarly, we could imagine asking whether or not "einsum" is responsible for the speedups here.

The specific case for "naive" is that we are not trying to arbitrarily make less efficient code -- this is the code we had when we first wrote the PyTorch versions. In good faith, we created something that was 100x slower than what we now know we can achieve.  I believe this is an outcome that is reflected in other use cases, and which it is worth being able to document.   

Of course, we could just show old and new graphs side by side, but that always leaves open the questions of calibration across runs and architectures.

awf:
Just looping back on this -- is the einsum fragment responsible for much of the speedup?

Issue #172, opened 2019-12-22
Out of memory checking


mikhailnikolaev:
The runners should check out of memory situations (e.g. catch OOM exceptions or perform apriori tests if it's possible) and return a special exit code if such a situation appears or will able to appear. The global runner in its turn should check this exit code and don't run tests of the bigger sizes if tests of the smaller sizes can't be performed due to OOM.

msdmkats:
Catching OOM is generally not feasible, e.g. in .NET when you do so you must terminate the app using `Environment.FailFast` method which does not allow setting exit code. It is also usually too late - by that time OS might have already killed some of the processes we need.

There's no way to do a priori tests in runners, because they don't know how much memory modules will need. Hypothetically, we could extend interfaces for modules with a method that would do such a check, but that would impose an extra and most often unnecessary burden on the author of the module.

I do, however, see, how a module over consuming memory and crashing the whole benchmarking process as a result might be problematic. As a guard measure I propose to add code to the global runner that would periodically (say, every 200 ms) would check, how much memory the current benchmark runner consumes, and kill it, if it consumes more than some configurable threshold.
This way we could stop runners before they hang the system or cause it to kill some important processes.

Issue #24, opened 2019-05-14
Some run-all.ps1 changes


tomjaguarpaw:
We should merge something like https://github.com/awf/autodiff/tree/toelli/run-all.  The commit that unredirects stderr should be tidied up a bit.

tomjaguarpaw:
Perhaps we can capture stderr, add the indent, and output it immediately.  I'm not sure how to do that in Powershell.

awf:
Oddly it's 2>file, e.g. 
``
& {write-host "una"; "conor"; write-error "fed"} 2>err.txt >out.txt
``

PR #177, opened 2019-12-24
Checking out of memory

mikhailnikolaev:
The global runner takes additional parameter that defines maximum RAM size that the runner can consume. If test takes more, then it will be killed. Also, guaranteed out of memory checking is performed the same way as guaranteed timeout checking.

This PR solves the issue #172.

### DIFF
--- a/ADBench/run-all.ps1
+++ b/ADBench/run-all.ps1
@@ -53,8 +53,9 @@ param(# Which build to test.
       # Kill the test after this many seconds 
       [double]$timeout=300,
 
-      # Kill the test if it consumes more than this many gigabytes of RAM
-      [double]$max_memory_amount_in_gb=8,
+      # Kill the test if it consumes more than this many gigabytes of RAM.
+      # If parameter value is less or equal to zero, RAM checking isn't performed.
+      [double]$max_memory_amount_in_gb=-1,
       
       # Where to store the ouput, defaults to tmp/ in the project root 
       [string]$tmpdir="",
@@ -204,7 +205,7 @@ function run_command ($indent, $outfile, $timeout, $cmd) {
 
                 $status = [RunCommandStatus]::Timeout
                 break
-            } elseif ($mem -ge $max_memory_amount_in_bytes) {
+            } elseif (($max_memory_amount_in_gb -gt 0) -and ($mem -ge $max_memory_amount_in_bytes)) {
                 $Process.Kill()
                 $mem = [math]::round($mem / (1024 * 1024 * 1024), 2)
                 Write-Host "${indent}Killed due to consuming $mem GB of operating memory"
@@ -265,7 +266,7 @@ if ($gmm_k_vals_param) { $gmm_k_vals = $gmm_k_vals_param }
 # define test sizes, to be sorted ascending
 sort_size_parameters
 
-# convert max memory amount parameter info to bytes
+# convert max memory amount parameter to bytes
 $max_memory_amount_in_bytes = $max_memory_amount_in_gb * 1024 * 1024 * 1024
 
 # Set tmpdir default
--- a/ADBench/run-all.ps1
+++ b/ADBench/run-all.ps1
@@ -51,7 +51,10 @@ param(# Which build to test.
       [double]$time_limit=10,  
       
       # Kill the test after this many seconds 
-      [double]$timeout=300,    
+      [double]$timeout=300,
+
+      # Kill the test if it consumes more than this many gigabytes of RAM
+      [double]$max_memory_amount=8,
       
       # Where to store the ouput, defaults to tmp/ in the project root 
       [string]$tmpdir="",
@@ -156,11 +159,12 @@ function mkdir_p($path) {
 }
 
 # Type of run command function returning value
-enum RunCommandStatus { Finished; Timeout }
+enum RunCommandStatus { Finished; Timeout; OutOfMemory }
 
 # Run command and (reliably) get output
 function run_command ($indent, $outfile, $timeout, $cmd) {
     write-host "Run [$cmd $args]"
+
     $ProcessInfo = New-Object System.Diagnostics.ProcessStartInfo
     $ProcessInfo.FileName = $cmd
     $ProcessInfo.RedirectStandardError = $true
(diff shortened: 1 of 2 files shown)


ADBench/run-all.ps1
```diff
@@ -51,7 +51,10 @@ param(# Which build to test.
       [double]$time_limit=10,  
       
       # Kill the test after this many seconds 
-      [double]$timeout=300,    
+      [double]$timeout=300,
+
+      # Kill the test if it consumes more than this many gigabytes of RAM
+      [double]$max_memory_amount=8,
```
tomjaguarpaw: I would suggest calling this `max_memory_amount_in_gb` so that it's clearer what the units are.

ADBench/run-all.ps1
```diff
@@ -169,26 +173,48 @@ function run_command ($indent, $outfile, $timeout, $cmd) {
...
+                $Process.Kill()
+                Write-Host "${indent}Killed after $timeout seconds"
+                Store-NonFatalError "Process killed after $timeout seconds`n[$cmd $args]"
+                create_timeout_file $outfile $timeout
+                return [RunCommandStatus]::Timeout
+            } elseif ($mem -ge $max_memory_amount_in_bytes) {
+                $Process.Kill()
+                Write-Host "${indent}Killed due to consuming $mem bytes of operating memory"
+                Store-NonFatalError "Process killed due to consuming $mem bytes of operating memory`n[$cmd $args]"
+                return [RunCommandStatus]::OutOfMemory
```
msdmkats: I think, displaying `$mem` in GB here would be more convenient for users.

ADBench/run-all.ps1
mikhailnikolaev: Done

ADBench/run-all.ps1
```diff
@@ -169,26 +173,48 @@ function run_command ($indent, $outfile, $timeout, $cmd) {
...
+            }
+        }
     }
+
+    $stdout = $Process.StandardOutput.ReadToEnd().Trim().Replace("`n", "`n${indent}stdout> ")
+    $stderr = $Process.StandardError.ReadToEnd().Trim().Replace("`n", "`n${indent}stderr> ")
+    $allOutput = "${indent}stdout> " + $stdout + "`n${indent}stderr> " + $stderr
+    Write-Host "$allOutput"
```
msdmkats: We should do this even (and especially) when we kill the process due to timeout or OOM.

ADBench/run-all.ps1
```diff
@@ -51,7 +51,10 @@ param(# Which build to test.
       [double]$time_limit=10,  
       
       # Kill the test after this many seconds 
-      [double]$timeout=300,    
+      [double]$timeout=300,
+
+      # Kill the test if it consumes more than this many gigabytes of RAM
+      [double]$max_memory_amount_in_gb=8,
```
msdmkats: Maybe increase this?
Do we know how many and which tests need more memory?

ADBench/run-all.ps1
mikhailnikolaev: Ok, I will change this. By the way, the last version of the runner didn't do it in the case of timeout as well

ADBench/run-all.ps1
mikhailnikolaev: Actually, I don't know. But I can increase it up to 16 GB (my machine performs all basical tests with this RAM size). Will it be enough?

ADBench/run-all.ps1
mikhailnikolaev: It makes sense

ADBench/run-all.ps1
msdmkats: Hmm.. Indeed. Strange.
Anyway, I think displaying these always is the correct thing to do. And almost trivial to implement.

tomjaguarpaw: (My opinion on this is that stderr and stdout should be captured, indented and reoutput *immediately* as I noted at https://github.com/awf/ADBench/issues/24#issuecomment-492312312.  It's actually quite annoying to have to wait for a run to finish before seeing its intermediate output.)

msdmkats: While I agree in principle, capture-indent-immediate reoutput is not as trivial, so it probably shouldn't a part of this PR.

ADBench/run-all.ps1
mikhailnikolaev: Done

ADBench/run-all.ps1
mikhailnikolaev: Initial changes are performed

ADBench/run-all.ps1
mikhailnikolaev: Another option is not to check RAM usage at all by default

msdmkats: Yes, I think that would be the right thing to do.
In my dream we would have defaulted this value to something like total system memory - 2 GB, but .NET Core does not have the necessary API at the moment.

mikhailnikolaev: RAM checking by default is turned off

docs/GlobalRunner.md
```diff
@@ -60,7 +60,7 @@ Parameters:
 
 - `-max_memory_amount_in_gb <Double>`
 
-    Kill the test if it consumes more than this amount of gigabytes of RAM.
+    Kill the test if it consumes more than this amount of gigabytes of RAM. Parameter value less or equal to zero, then RAM checking isn't performed.
```
msdmkats: I think, you missed an `if`.

mikhailnikolaev: Fixed

ADBench/run-all.ps1
```diff
@@ -51,7 +51,11 @@ param(# Which build to test.
       [double]$time_limit=10,  
       
       # Kill the test after this many seconds 
-      [double]$timeout=300,    
+      [double]$timeout=300,
+
+      # Kill the test if it consumes more than this many gigabytes of RAM.
+      # If parameter value is less or equal to zero, RAM checking isn't performed.
+      [double]$max_memory_amount_in_gb=-1,
```
tomjaguarpaw: Does PowerShell have something you could use in the same way that one would use `None` in Python?  i'm not particularly comfortable storing "no limit" in-band as a negative number.

awf:
There is $null, which you would need to check using “$x -eq $null”.  It’s a little tricky as PowerShell does so much null-conversion anyway, so there’s an argument for the sentinel too.

ADBench/run-all.ps1
msdmkats: While '0 means unlimited' is a quite traditional convention used in many programs, there is, as 
@awf mentioned, `$null`. However, if we're going for non-numbers, I'd like to add `[double]::PositiveInfinity` to consideration. It has a very direct "no limit" meaning and allows to use the same code for both limited and unlimited scenarios.

awf: Yes, I think that's the best solution.

tomjaguarpaw: I agree. It would allow us to remove the check `$max_memory_amount_in_gb -gt 0` completely.

mikhailnikolaev: I've added using positive infinity instead of negative numbers. So, I think, this issue is solved

PR #178, opened 2019-12-24
Dmkats/buildfix

### LINKED ISSUE #176: TensorFlow missing in Windows PR tests
When running the "ADBench PR Build Win Test" it complains that TensorFlow is missing, for example

```
======================================================================
ERROR: test_simple_objective_runs_multiple_times (__main__.PythonModuleCommonHandTests)
Checks if objective can be calculated multiple times
----------------------------------------------------------------------
Traceback (most recent call last):
  File "D:/a/1/s/test/python/modules/common/hand_tests.py", line 174, in setUp
    self.test = module_load(module_path)
  File "D:\a\1\s\src\python\runner\ModuleLoader.py", line 21, in module_load
    spec.loader.exec_module(test)
  File "<frozen importlib._bootstrap_external>", line 783, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "D:\a\1\s\src\python\modules\TensorflowGraph\TensorflowGraphHand.py", line 8, in <module>
    import tensorflow as tf
ModuleNotFoundError: No module named 'tensorflow'
...
The following tests FAILED:
	172 - PythonBaModuleCommonTests (Failed)
	173 - PythonGmmModuleCommonTests (Failed)
	174 - PythonLstmModuleCommonTests (Failed)
	175 - PythonHandModuleCommonTests (Failed)
```

Should it be `pip install`ed, or obtained some other way?

https://msrcambridge.visualstudio.com/Knossos/_build/results?buildId=48452&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=f33089cd-d3fe-578c-2c5e-4ee560e10a0d&l=2139

msdmkats:
Fixes #176 
Build: https://msrcambridge.visualstudio.com/Knossos/_build/results?buildId=48485

The problem was that cmake's find_package found Python 3.8, instead of the default 3.7.
Python 3.8 is not yet supported by all pip packages.

### DIFF
--- a/adobuilds/pr-win.yml
+++ b/adobuilds/pr-win.yml
@@ -21,6 +21,14 @@ jobs:
         architecture: x64
 
     - script: |
+        python -m pip install --upgrade pip
+      displayName: Upgrade pip
+
+    - script: |
+        python -m pip install pip setuptools>=41.0.0
+      displayName: Install setuptools
+
+    - script: |
         choco install --limit-output julia
       displayName: Install Julia
 
--- a/adobuilds/pr-win.yml
+++ b/adobuilds/pr-win.yml
@@ -15,6 +15,7 @@ jobs:
 
   steps:
     - task: UsePythonVersion@0
+      name: UsePython
       inputs:
         versionSpec: 3.7
         addToPath: true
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
Well done for tracking that down.  The way that ADO provides versions of things that you didn't ask for can be very annoying.

msdmkats:
It was not ADO, it was CMake. Both python versions (and many more, in fact) are present on the build machine, there's a task that we use that makes the chosen version the default one.

But then CMake uses some non-obvious algorithm and picks the wrong one, unless forced to do otherwise.

tomjaguarpaw:
> Both python versions (and many more, in fact) are present on the build machine, there's a task that we use that makes the chosen version the default one.

Yes, I know, and I don't like it.  It makes these sorts of bugs more likely to occur.  Ideally the only software installed on the build machine would be that which is explicitly chosen in the build definition. I'm sure that would not be straightforward to achieve though.  Anyway, this is rather tangential ...

Issue #25, opened 2019-05-15
Export data file for the graphs


tomjaguarpaw:
It would be nice if the plotting process exported a data file (perhaps JSON) containing the all the datapoints created by the training run.  Ideally the plotting code could be used to replot the graphs from the data file.

tomjaguarpaw:
@mikhailnikolaev @msdmkats are the data files that the graphs are plotted from currently uploaded to the web server?

mikhailnikolaev:
> @mikhailnikolaev @msdmkats are the data files that the graphs are plotted from currently uploaded to the web server?

No, they aren't. Actually, only PNG and HTML files (i.e. static and plotly graphs) are uploaded to the blob container in the current version. Moreover, the plotting process still doesn't collect all time info into the one JSON file.

mikhailnikolaev:
@tomjaguarpaw, I think, we can close this issue due to the PR #183 and #186 (that are parts of #180). What do you think?

tomjaguarpaw:
I would say either

1. Don't close this until JSON export functionality is added to the Azure CI and Batch job, or
2. Close this and create a new issue for the case above

PR #180, opened 2019-12-25
Time info collecting

mikhailnikolaev:
The `plot-graphs.py` script now collects all time information, that was used for creating graphs, into the single JSON file. Then the script can use such a file for creating graphs without any information generated by the global runner. For specifying such a behaviour the argument `--use-file <file>` is used.

This PR is linked to issue #25.


ADBench/plot_graphs.py
```diff
@@ -387,14 +422,16 @@ def generate_graph(figure_idx, graph_function_type):
...
 
-    build_type = graph[0]
-    maybe_test_size = graph[2:]
-    (graph_name, graph_save_location) = graph_data(build_type, objective, maybe_test_size, function_type)
+    (graph_name, graph_save_location) = graph_data(figure_info)
 
     # Setup graph attributes
+    xlabel = "No. independent variables"
+    if "hand" == figure_info.objective or "hand" in figure_info.maybe_test_size:
+        xlabel = "No. correspondencies"
```
awf: correspondencies -> correspondences

mikhailnikolaev: Hmmm... interesting, it looks like this misprint has been being here for a long time)
Fixed.

ADBench/plot_graphs.py
```diff
@@ -446,19 +483,88 @@ def generate_graph(figure_idx, graph_function_type):
...
+    figure_info.objective = data["objective"]
+    figure_info.maybe_test_size = data["test_size"]
+    figure_info.function_type = data["function_type"]
+
+    return sorted_vals_by_tool, figure_info
+
 def main():
     print_messages()
 
+    plot_data = []      # holds all the data that is plotted in the graph
```
tomjaguarpaw: At the moment the logic for generating plot data is coupled to graph generation. They should be separated.  The logic for this section should look more like

1. `if use_file` then get `plot_data` from file, otherwise generate `plot_data` by looping over all graphs
2. Unconditionally generate the graphs from `plot_data`

ADBench/plot_graphs.py
```diff
@@ -212,21 +234,26 @@ def print_messages():
...
+that is used for graph creating.
 CMD arguments:
     --save
-            if specified then script saves produced graphs to
+            if specified then the script saves produced graphs to
             {static_out_dir}
 
     --plotly
-            if specified then script saves graphs in plotly format to
+            if specified then the script saves graphs in plotly format to
```
tomjaguarpaw: Different people have different preferences regarding this, but FYI my personal preference is if typo-changes are submitted in separate PRs.  The difficulty for me of reviewing a PR is O(n^2) in the number of lines changed (I think about all possible new interactions between pairs of lines changed) so submitting independent changes in independent PRs picks up a significant performance improvement.

tomjaguarpaw: (This is an observation only, not a request for a change to this PR!)

ADBench/plot_graphs.py
```diff
@@ -56,34 +58,54 @@
...
-all_files = [path for path in utils._scandir_rec(in_dir) if TIMES_SUBSTRING in path[-1]]
-all_graphs = [path.split("/") for path in set(["/".join(path[:-2]) for path in all_files])]
-function_types = ["objective ÷ Manual", "objective", "jacobian", "jacobian ÷ objective"]
-all_graphs = [(path, function_type) for function_type in function_types for path in all_graphs]
-all_graph_dict = {}
+if not use_file:
+    all_files = [path for path in utils._scandir_rec(in_dir) if TIMES_SUBSTRING in path[-1]]
+    all_graphs = [path.split("/") for path in set(["/".join(path[:-2]) for path in all_files])]
+    function_types = ["objective ÷ Manual", "objective", "jacobian", "jacobian ÷ objective"]
+    all_graphs = [(path, function_type) for function_type in function_types for path in all_graphs]
```
tomjaguarpaw: These variables are needed only in `main` under the `else` branch of an `if use_file` (`all_files` is used indirectly in the call of `get_sorted_vals_by_tool`) so these bindings should be moved there (and `all_files` passed as an argument to `get_sorted_vals_by_tool`).

ADBench/plot_graphs.py
```diff
@@ -56,34 +58,54 @@
...
+# File with input data if specified
+if use_file:
+    arg_idx = sys.argv.index("--use-file") + 1
+    input_file_path = sys.argv[arg_idx]
+
 # Folders
 adbench_dir = os.path.dirname(os.path.realpath(__file__))
 ad_root_dir = os.path.dirname(adbench_dir)
-in_dir = os.path.join(ad_root_dir, "tmp")
+in_dir = os.path.join(ad_root_dir, "tmp") if not use_file else os.path.dirname(input_file_path)
```
tomjaguarpaw: If `use_file` is false then `in_dir` should never be used, is that not the case?  If it is the case then I would suggest setting `in_dir` to `None` to indicate that (pending a future refactoring where we only ever assign `in_dir` when we know it will be used).

tomjaguarpaw:
To expand a bit, this module now performs (or should perform) two conceptually distinct tasks:

1. Generate and collate graph data from a collection of files scattered around the filesystem and put it into a single object
2. Plot graphs from the single object

(additionally it deals with loading and saving the single object to the filesystem, but that's a triviality)

It's very important that these two conceptually distinct tasks are kept distinct in the implementation too (i.e. that they be decoupled).

mikhailnikolaev:
I think I should split this PR into the following parts:

- [ ] Separating get values and drawing graph.
- [ ] Generating and saving plot data. Using plot data for creating graphs.
- [ ] Using specified file for creating graphs. Also, performing changes, requested in comments https://github.com/awf/ADBench/pull/180#discussion_r361458314 and https://github.com/awf/ADBench/pull/180#discussion_r361457678
- [ ] Doc string and output reference fixing/changing.
- [ ] Documentation changes.

@tomjaguarpaw, what do you think?

tomjaguarpaw:
For me, splitting it would help a lot and make it easier to review.  Thanks!

PR #183, opened 2019-12-27
Collecting plot data [part of #180]

mikhailnikolaev:
Plot script now collects plot data and then uses it for creating graphs. Also, it saves plot data to the JSON file.

This PR contains a part of changes from the closed PR #180. It is the second part of the list, described here https://github.com/awf/ADBench/pull/180#issuecomment-569066919

### DIFF
--- a/ADBench/plot_graphs.py
+++ b/ADBench/plot_graphs.py
@@ -337,12 +337,12 @@ def values_and_styles(sorted_vals_by_tool):
 
         yield item, style[0: 2], display_name
 
-def generate_graph(figure_info, sorted_vals_by_tool):
+def generate_graph(idx, figure_info, sorted_vals_by_tool):
     '''Generates the graph for the given figure.
     
     Args:
+        idx: index of the figure.
         figure_info (named tuple): information of the figure.
-            idx: index of the figure.
             build_type: the type of the tool build.
             objective: the name of the objective, the graph is plotted.
             maybe_test_size: test size or empty string if the test can not have
@@ -353,7 +353,7 @@ def generate_graph(figure_info, sorted_vals_by_tool):
     '''
 
     # Create figure
-    figure = pyplot.figure(figure_info.idx, figsize=figure_size, dpi=fig_dpi)
+    figure = pyplot.figure(idx, figsize=figure_size, dpi=fig_dpi)
 
     handles, labels = [], []
     non_timeout_violation_x, non_timeout_violation_y = [], []
@@ -505,7 +505,7 @@ def extract_vals_and_figure_info_from_plot_data(data):
 
     figure_info = namedtuple(
         "figure_info",
-        "idx, build_type, objective, maybe_test_size, function_type"
+        "build_type, objective, maybe_test_size, function_type"
     )
 
     figure_info.build_type = data["build"]
@@ -525,8 +525,7 @@ def main():
     print("\nGenerating graphs...\n")
     for (figure_idx, data) in enumerate(plot_data, start=1):
         sorted_vals_by_tool, figure_info = extract_vals_and_figure_info_from_plot_data(data)
-        figure_info.idx = figure_idx
-        generate_graph(figure_info, sorted_vals_by_tool)
+        generate_graph(figure_idx, figure_info, sorted_vals_by_tool)
 
     print(f"\nPlotted {figure_idx} graphs")
 
--- a/ADBench/plot_graphs.py
+++ b/ADBench/plot_graphs.py
@@ -334,7 +334,7 @@ def values_and_styles(sorted_vals_by_tool):
 
         display_name = utils.format_tool(tool) if len(style) == 2 else style[2]
 
-        yield item.values(), style[0: 2], display_name
+        yield item, style[0: 2], display_name
 
 def generate_graph(idx, data):
     '''Generates the graph from the given data.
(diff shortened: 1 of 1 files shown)


ADBench/plot_graphs.py
```diff
@@ -459,32 +459,85 @@ def generate_graph(figure_info, sorted_vals_by_tool):
...
+    plot_data = get_plot_data(all_graphs)
 
-        (graph, figure_info.function_type) = t
-        figure_info.build_type = graph[0]
-        figure_info.objective = graph[1]
-        figure_info.maybe_test_size = graph[2] if len(graph) == 3 else ""
+    # Loop through each of graphs to be created
+    print("\nGenerating graphs...\n")
+    for (figure_idx, data) in enumerate(plot_data, start=1):
+        sorted_vals_by_tool, figure_info = extract_vals_and_figure_info_from_plot_data(data)
```
tomjaguarpaw: It seems to me that `figure_idx` should be passed into `extract_vals_and_figure_info_from_plot_data` and set within that function, rather than being set afterwards.  What do you think?

mikhailnikolaev: Actually, I think, it would be even better if `idx` were not a part of figure info at all. It seems like the best option

mikhailnikolaev: Changed

tomjaguarpaw: Do you mean instead to pass it in directly as an argument to `generate_graph`?  If so then I agree.  That sounds like an even better idea.

mikhailnikolaev:
> This looks great, but is there any way that the same data structure can be used for both serialisation and generating the graph? Is it possible to use a singled named tuple and write that out as JSON?

Will it be more convenient if the `plot_data,json` file has the following form:
```json
[
    {
        "figure_info": {
            "build": build,
            "objective": objective,
            "function_type": function_type,
            "test_size": test_size
        },
        "values": [ ... ]
]
```
Thus the script can use `figure_info` subdictionary instead of named tuple

ADBench/plot_graphs.py
mikhailnikolaev: Yes, that's what I meant

tomjaguarpaw:
Yes, I think that would clear up the biggest differences. It would be good if we could avoid preprocessing `sorted_vals_by_tool` too.

mikhailnikolaev:
> Yes, I think that would clear up the biggest differences. It would be good if we could avoid preprocessing `sorted_vals_by_tool` too.

Actually, I have the idea to use only items of the `plot_data` array everywhere in the script. I mean, we can even avoid such a term as `sorted_vals_by_tool` and use only dictionaries of the form
```json
{ "figure_info": { ... }, "values": { ... } }
```
where "values" is a kind of `sorted_vals_by_tools` and 'figure_info" is `figure_info`. Thus, we can avoid any additional serialization actions except `json.dump`.  But I don't know is it a good idea or there is a more pythonic way. What do you think?

tomjaguarpaw:
Ideally the form that we store as JSON would be as close as possible to the form that is used by `generate_graph`. It seems like it would be nice not to unwrap and rewrap `sorted_vals_by_tool`.  What do you think?

tomjaguarpaw:
OK, I think your idea sounds good.  Let's give it a go!

mikhailnikolaev:
> Ideally the form that we store as JSON would be as close as possible to the form that is used by `generate_graph`. It seems like it would be nice not to unwrap and rewrap `sorted_vals_by_tool`. What do you think?

This will be one of the positive corollaries of using dictionary everywhere instead of `sorted_vals_by_tools`

mikhailnikolaev:
> OK, I think your idea sounds good. Let's give it a go!

Should I do another PR, or I can add commits into this one?

tomjaguarpaw:
> Should I do another PR, or I can add commits into this one?

I think best to add commits into this one and see what it looks like.  Then we can decide what to do.

tomjaguarpaw:
> I think best to add commits into this one and see what it looks like. Then we can decide what to do.

In fact, you can do whichever you prefer.

ADBench/plot_graphs.py
```diff
@@ -325,7 +324,7 @@ def values_and_styles(sorted_vals_by_tool):
 
     next_default = 0
     for item in sorted_vals_by_tool:
-        tool = item[0]
+        tool = item["tool"]
```
tomjaguarpaw: Nice. This is much clearer now!

ADBench/plot_graphs.py
```diff
@@ -335,32 +334,39 @@ def values_and_styles(sorted_vals_by_tool):
...
-    figure = pyplot.figure(figure_info.idx, figsize=figure_size, dpi=fig_dpi)
+    figure = pyplot.figure(idx, figsize=figure_size, dpi=fig_dpi)
 
     handles, labels = [], []
     non_timeout_violation_x, non_timeout_violation_y = [], []
     additional = []
 
     # Plot results
-    for ((tool, n_vals, t_vals, violations), style, disp_name) in values_and_styles(sorted_vals_by_tool):
+    for ((tool, t_vals, n_vals, violations), style, disp_name) in values_and_styles(data["values"]):
```
tomjaguarpaw: I think it might be clearer if you return `items` from `values_and_styles` and then explicitly index into that dictionary when you call `label_and_handle` and `together_and_additionals`.

tomjaguarpaw:
This is improving a lot.

ADBench/plot_graphs.py
mikhailnikolaev: Done

ADBench/plot_graphs.py
mikhailnikolaev: Yes, I think the same way)

PR #185, opened 2019-12-27
IO paths creating refactoring [part of #180]

mikhailnikolaev:
This PR contains a part of changes from the closed PR #180. It is the fourth part of the list, described here https://github.com/awf/ADBench/pull/180#issuecomment-569066919

### DIFF
--- a/ADBench/plot_graphs.py
+++ b/ADBench/plot_graphs.py
@@ -200,7 +200,7 @@ def draw_vertical_lines(vals_by_tool):
     for n in all_n_vals:
         pyplot.axvline(n, ls = '-', color = "lightgrey", zorder = 0.0, lw = 0.5)
 
-def print_messages(out_dir):
+def print_messages():
     '''Prints messages and exits the program if --help were specified'''
 
     if do_help:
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
This looks like exactly the right thing to do, but should we pass `static_output_dir` and `plotly_output_dir` to `print_messages`?

mikhailnikolaev:
> This looks like exactly the right thing to do, but should we pass `static_output_dir` and `plotly_output_dir` to `print_messages`?

Actually, I think, that the refernce should be static. I know that I was the one who made it non-static, but now I've changed my mind. Firstly, all info about output and input dirs the user gets in the console when the script working. Secondly, in the future change the output dir is variative, I mean its path depend on giving (or not) `--use-file` argument. So, my proposal is to make the reference in the way it is now. Or do you disagree with me?

tomjaguarpaw:
I think you are correct because static output is simpler and it is not critical to have it dynamic.  Let's keep things as simple as we can.

mikhailnikolaev:
> This is great.

Thanks)

PR #186, opened 2019-12-27
Using specified plot data file [part of #180]

mikhailnikolaev:
Now the script can use specified plot data file instead of the global runner output files.

This PR contains a part of changes from the closed PR #180. It is the fifth part of the list, described here https://github.com/awf/ADBench/pull/180#issuecomment-569066919

### DIFF
--- a/ADBench/plot_graphs.py
+++ b/ADBench/plot_graphs.py
@@ -219,6 +219,11 @@ CMD arguments:
             screen. Note, that this is default option if --save or
             --plotly are not defined.
 
+    --use-file <file>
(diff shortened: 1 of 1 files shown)


mikhailnikolaev:
> It looks like `get_out_dir_and_plot_data` should be two separate functions, because it does one of two completely independent things depending on whether `use_file`.

Not exactly. In the case of `--use-file` we need to extract the value of the cmd argument both for creating output dir and getting plot data. Now we do it only once. I thought, it was a kind of optimization, but maybe it's better to do it twice instead of creating complicated code

PR #195, opened 2020-09-09
Recommend venv

cgravill:
As part of the install pip will currently changes the Python environment including replacing existing library versions without warning. If someone is using Conda environments, for example, this behaviour may be quite surprising.


tomjaguarpaw (approved):
Thanks Colin

PR #196, opened 2020-09-09
Rely on environment variables to find Python

cgravill:
When CMake is locating Python on Windows https://cmake.org/cmake/help/latest/module/FindPython3.html it defaults to searching the registry first. This is awkward behaviour on e.g. GitHub Actions where it has a later version of Python available but wouldn't be used if calling e.g. `python -V`.

I think it's closer to expectations to respect the environment variables instead where we want to use a precise version for benchmark comparability. That matches the current documentation which instructs put the binaries on the PATH:

https://github.com/microsoft/ADBench/blob/191c5244e7a815984d302ccbc48372137235a9dd/docs/BuildAndTest.md#prerequisites

### DIFF
--- a/ADBench/CMakeLists.txt
+++ b/ADBench/CMakeLists.txt
@@ -13,6 +13,7 @@ endfunction()
 JOIN("${GMM_D_VALS}" ", " GMM_D_STR)
 JOIN("${GMM_K_VALS}" ", " GMM_K_STR)
 
+set (Python3_FIND_REGISTRY NEVER)
 find_package(Python3 COMPONENTS Interpreter)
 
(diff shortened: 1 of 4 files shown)


tomjaguarpaw:
Does this mean that it ignores the registry entirely?

cgravill:
With this proposed new setting yes. It completely ignores the registry.

A more friendly option would be to use the LAST which first uses environment variables would then fall back on the registry if there's nothing on PATH.

tomjaguarpaw (approved):
I'm perfectly happy with ignoring the registry entirely.  The less implicit configuration the better in my opinion!

awf:
Yep, great fix.  Ouch!

PR #197, opened 2020-09-09
Get automated builds working again 

cgravill:
I've ported https://github.com/microsoft/ADBench/blob/9883bcdf06c27465ce7bd94ad4cff264a62f98ec/adobuilds/pr-win.yml over to GitHub Actions. That lets the build be more integrated rather than potentially calling into an external system (which is also not currently active).

There are a fair number of hard-coded values as before, we might want to look at extracting them to share with scripts but I'm keen to get a CI build active then expand to more of the benchmarks. This currently just does: `("Finite", "Manual", "PyTorch")`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,81 @@
+name: ManualAndPyTorchBenchmarks
+
+on: [pull_request]
+
+env:
+  # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
+  BUILD_TYPE: Release
+
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


awf (approved):
This looks excellent, thanks.

tomjaguarpaw:
Cool! The plots generated for this particular run are at https://github.com/microsoft/ADBench/actions/runs/246611609

tomjaguarpaw (approved):

toelli-msft (approved):

cgravill:
Yes, it'd be great when I can figure out how to attach the benchmark plots inline to the PR but it's a nice start having them accessible through the checks UI.

PR #198, opened 2020-10-08
Lift the GMM repetitions into TorchScript

cgravill:


src/python/modules/TorchScript/TorchScriptGMM.py
```diff
@@ -9,7 +11,16 @@
 from shared.GMMData import GMMInput, GMMOutput
 from modules.TorchScript.gmm_objective import gmm_objective
 
-
+# TorchScript doesn't currently support * argument unpacking so make an explicit set of arguments.
+@torch.jit.script
+def calculate_objective_explicit(times:int, alphas, means, icf, x, wishart_gamma, wishart_m):
```
awf, line 16: I feel the real difference in this function is that it's jitted.  
Suggest calling it `calculate_objective_jit_script`?
It also does the unpacking, but that feels secondary.

awf (changes requested):

src/python/modules/TorchScript/TorchScriptGMM.py
cgravill: Hmm I can see this either way, it's the `@torch.jit.script` annotation that's making it jitted, but I agree from the call site what you care about is that it's marked as TorchScript. Another way to do this is not use the annotatio  and instead call `torch.jit.script(calculate_objective_explicit)` where presumably we wouldn't add a `_script` suffix. Should we be marking all the method with the TorchScript annotation also with `_script`? This said, in this context the sole purpose of the function is to make it TorchScript compatible so happy to go with your suggestion if it clarifies the code for you.

awf: > `torch.jit.script(calculate_objective_explicit)` 
Do I understand this right -- one could write
```
   calculate_objective_ts = torch.jit.script(calculate_objective)
```  
I really like this idea , and we can do it outside the timing loop to be super explicit.  

I also half-like the idea of marking every annotated method -- we would need to use a shorter string, e.g. _ts -- but on balance I agree that that feels odd (as it's just about the codegen).

awf (approved):

src/python/modules/TorchScript/TorchScriptGMM.py
cgravill: Yes, the syntax we could use:

```python
        calculate_objective_explicit_as_script_function = torch.jit.script(calculate_objective_explicit)
        self.objective = calculate_objective_explicit_as_script_function(times, *self.inputs, *self.params) # Run with the TorchScript interpreter
```

In some ways I prefer that as the decorator isn't making it very explicit at the call site that you're using the TorchScript interpreter. However, I haven't generally done that for a few reasons:

- It's the not the main documented way in PyTorch documentation, they encourage the use of the decorator on functions
- If you have a chain of function A, B, C and there's a programming error in C just converting A gives worse errors than marking for eager conversion A, B, C. That's a lot easier with decorators than maintaining calls to convert each function

awf: Ah yes I see, and then it seems odd to not decorate the topmost call.  In that case, I'm inclined to go back to the original PR, but I'm still not super keen on the "_explicit" naming -- I do think that a "_ts" feels appropriate in this case.

awf: [And the distinction between splatted and unsplatted args doesn't feel important to stress, IMO]

cgravill: Thanks, changed the name to _ts. I'll merge after the build complete.

PR #199, opened 2020-10-30
Update the .NET Core module to target .NET Core 3.1

cgravill:
In order to update https://diffsharp.github.io/ to 1.0 we need a higher version of .NET Core, I've moved up to the next LTS version, 3.1. I'm putting this change in now as a good incremental step to match other tools.


awf (approved):
Lgtm thanks!

cgravill:
Based on local runs, I'm not seeing any major performance difference for GMM.

Before change:
GMM (10k)  Jacobian ÷ objective  - Release Graph

After change:
GMM (10k)  Jacobian ÷ objective  - Release Graph

PR #200, opened 2020-11-04
Devops build tidy

cgravill:
Ensure a matplotlib compatible with plotly
Get a current .NET Core SDK for builds to match .NET projects.
Reduce time spent on dependencies

### DIFF
--- a/ADBench/azure-devops-ci.ps1
+++ b/ADBench/azure-devops-ci.ps1
@@ -4,8 +4,8 @@
 choco install --limit-output julia
 julia -e 'import Pkg; Pkg.add.([\"SpecialFunctions\", \"ForwardDiff\"])'
 
-python -m pip install --user matplotlib plotly numpy scipy autograd
-python -m pip install torch==1.2.0+cpu torchvision==0.4.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
+# python -m pip install --user matplotlib plotly numpy scipy autograd
+# python -m pip install torch==1.2.0+cpu torchvision==0.4.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
 
 # I'm not sure that setting up vcvars is necessary in Azure Dev Ops.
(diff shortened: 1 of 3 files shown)


ADBench/azure-devops-ci.ps1
```diff
@@ -4,8 +4,8 @@
 choco install --limit-output julia
 julia -e 'import Pkg; Pkg.add.([\"SpecialFunctions\", \"ForwardDiff\"])'
 
-python -m pip install --user matplotlib plotly numpy scipy autograd
-python -m pip install torch==1.2.0+cpu torchvision==0.4.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
+# python -m pip install --user matplotlib plotly numpy scipy autograd
+# python -m pip install torch==1.2.0+cpu torchvision==0.4.0+cpu -f https://download.pytorch.org/whl/torch_stable.html
```
tomjaguarpaw, line 8: I can't understand how numpy, scipy, autograd, torch and torchvision get installed without these lines.  `ADBench/requirements.txt` only contains matplotlib and plotly.

tomjaguarpaw: Oh, is it `python/modules/PyTorch/requirements-....txt` that makes sure they are present?

cgravill: Yes, there's a split here. The first line "matplotlib..." are purely for the charts, the second "torch..." are for the benchmarks. Both are now not needed here as far as I can tell, the charts are handled by:

https://github.com/microsoft/ADBench/blob/87af0219a568807f8432754688ceb636efac12c6/ADBench/requirements.txt

and per benchmark ones multiple requirements files: shared, cuda, and noncuda ones.

The effect of the above was to install modules, which either weren't relevant or immediately upgraded. I couldn't find anywhere that was reliant on this script installing the packages.

toelli-msft (approved):

PR #205, opened 2021-04-28
Print reason when dlopen fails.

athas:
I don't know the equivalent of `dlerror()` on Windows, but it shouldn't be difficult to support if someone does.


toelli-msft:
Thanks! This looks fine to me but can someone with more C++ experience comment? @dcrc2 @awf?

dcrc2 (approved):
Seems fine to me. I'm happy with not trying to add the Windows version in this PR; my experience is that the Windows error messages for LoadLibrary aren't very helpful anyway.

athas:
Rebased.

PR #206, opened 2021-04-28
Create output directories if they do not exist.

athas:
Without this, the script simply fails if the directories do not exist.


toelli-msft (approved):
Great idea, thank you! Maybe @dcrc2 or @cgravill should take a quick look too.

cgravill (approved):
Looks good to me too

PR #207, opened 2021-06-01
Make ADBench work on macOS.

athas:
In most cases the Linux code also works on macOS, because we're only
using POSIX APIs.


toelli-msft (approved):
Thanks! Looks good to me. @awf or @dcrc2 can you double check?

dcrc2 (approved):
Looks good to me

awf (approved):
Yep, concur.

PR #208, opened 2021-06-01
Explicitly ask for C++14 support.

athas:
This is related to #207 because macOS's default C++ compiler assumes it's still ten years ago.


toelli-msft (approved):
Thanks! Looks good to me. @awf or @dcrc2 can you double check?

athas:
Woops, messed up with a rebase somehow.  This change is still necessary to make ADBench work out-of-the-box on macOS (but that should be the last one).

dcrc2 (approved):

PR #209, opened 2021-06-01
This should be a bitwise or, not a logical or.

athas:
Remarkably it still worked before, but who knows what it did (and the compiler complained).

### DIFF
--- a/src/cpp/runner/ModuleLoader.cpp
+++ b/src/cpp/runner/ModuleLoader.cpp
@@ -18,7 +18,7 @@ ModuleLoader::ModuleLoader(const char* file_path)
 #ifdef _WIN32
     module_ptr_ = LoadLibraryA(file_path);
 #elif defined(__linux__) || defined(__APPLE__)
-    module_ptr_ = dlopen(file_path, RTLD_NOW || RTLD_LOCAL);
+    module_ptr_ = dlopen(file_path, RTLD_NOW | RTLD_LOCAL);
 #endif
     if (module_ptr_ == nullptr) {
         throw runtime_error(("Can't load library " + std::string(file_path)).c_str());


toelli-msft:
Thanks!

My `/usr/include/x86_64-linux-gnu/bits/dlfcn.h` contains the following, so  `RTLD_NOW || RTLD_LOCAL == 1 == RTLD_LAZY | RTLD_LOCAL` (probably). I guess we were resolving symbols lazily by accident.


```h
/* The MODE argument to `dlopen' contains one of the following: */
#define RTLD_LAZY       0x00001 /* Lazy function call binding.  */
#define RTLD_NOW        0x00002 /* Immediate function call binding.  */
#define RTLD_BINDING_MASK   0x3 /* Mask of binding time value.  */
#define RTLD_NOLOAD     0x00004 /* Do not load the object.  */
#define RTLD_DEEPBIND   0x00008 /* Use deep binding.  */

/* If the following bit is set in the MODE argument to `dlopen',
   the symbols of the loaded object and its dependencies are made
   visible as if the object were linked directly into the program.  */
#define RTLD_GLOBAL     0x00100

/* Unix98 demands the following flag which is the inverse to RTLD_GLOBAL.
   The implementation does this by default and so we can define the
   value to zero.  */
#define RTLD_LOCAL      0
```

toelli-msft:
@awf or @dcrc2, probably worth having one of you double check this.

athas:
Rebased on top of master to avoid conflicts.

awf:
Thanks @athas!  @toelli-msft can we merge this?

toelli-msft:
Yes, we can merge. It had a test failure that I think may have been transient. I've kicked off another run.

PR #210, opened 2021-06-25
Add missing include in finite.h

tomsmeding:
On my machine, the C++ code failed to build because of the missing include in this PR. Apparently, `std::numeric_limits` was indirectly included/defined in one of the other header files on older GCC versions; I'm running GCC 11.1.0.


toelli-msft:
Thanks! @dcrc2 or @awf can you check this for me?

toelli-msft (approved):

dcrc2 (approved):
Yes this is absolutely correct, thanks.

